PR #133, opened 2026-06-18
Rescue: preserve waterui-math component from deleted branch

lexoliu:
## Rescue: `waterui-math` component from deleted branch

The branch `feat/math-font-name-source` was deleted during a branch cleanup. Before deletion it was found to contain a **complete, ~3080-line `waterui-math` formula-rendering component** plus an example that existed **nowhere else** — neither on `dev` nor on `main` (`main` and `dev` have fully disjoint histories, so a direct cherry-pick was impossible).

This PR preserves that unique work on a `dev`-based branch so it is not lost.

### What's included (verbatim from the deleted branch)
- `components/math/` — the `waterui-math` crate (`src/lib.rs` 3080 lines, `Cargo.toml`, `README.md`). Vello-based math/LaTeX/MathML/Typst formula rendering. Declares `STIX Two Math` + `Noto Sans Math` font assets.
- `examples/math/` — the math rendering example.

### Integration still required before merge (NOT done here)
This is content preservation, not a finished integration. Before merging:
1. **Relocate to `dev`'s restructured tree.** `dev` now groups components by category — the crate should move to `components/visual/math/` (joins the `components/visual/*` workspace glob) and the example to the appropriate `examples/showcase/...` bucket. They are intentionally placed at the legacy flat paths here (`components/math`, `examples/math`), which `dev`'s member globs do **not** match, so dev's build is unaffected and CI stays clean until wired in.
2. **Wire the workspace.** Add `waterui-math` to `[workspace.dependencies]` (deps `vello`/`parley`/`kurbo`/`peniko`/`waterui-graphics`/`-layout`/`-text`/`-str` already exist on `dev`) and the root `waterui` re-export if desired.
3. **Resolve API drift.** The component was written against the deleted branch's (pre-restructure) versions of `waterui-graphics` / `-layout` / `-text`. Those APIs have advanced on `dev`; expect compile fixes.

Source of truth: commit `d6122a85` ("feat(math): add waterui-math component and integrate formula rendering").

### DIFF
--- a/components/math/Cargo.toml
+++ b/components/math/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "waterui-math"
+version = "0.1.0"
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+readme = "README.md"
+repository.workspace = true
+description = "Math formula rendering for WaterUI using Vello"
+keywords = ["math", "latex", "mathml", "typst", "vello"]
+categories = ["gui"]
+
+[dependencies]
+waterui-core.workspace = true
+waterui-graphics.workspace = true
+waterui-layout.workspace = true
+waterui-str.workspace = true
+waterui-text.workspace = true
+parley.workspace = true
+vello.workspace = true
+kurbo.workspace = true
+peniko.workspace = true
+ttf-parser = "0.25"
+
+[lints]
+workspace = true
+
+[package.metadata.waterui.assets]
+
+[[package.metadata.waterui.assets.font]]
+name = "STIX Two Math"
+
+[[package.metadata.waterui.assets.font]]
+name = "Noto Sans Math"
--- a/components/math/README.md
+++ b/components/math/README.md
@@ -1,0 +1,38 @@
+# waterui-math
+
+GPU math formula rendering for WaterUI with Vello.
+
+## Input formats
+
+- LaTeX math subset
+- MathML Core subset
+- Typst math subset
+
+## Rendering model
+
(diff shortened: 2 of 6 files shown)


lexoliu:
Superseded by #226, which landed `waterui-math` at `components/visual/math/` — merged into dev as 659f6f2a9, closing #213.

Merging this now would add a second math crate at the legacy flat path `components/math/`, alongside a working one.

The code this rescued was audited in full on 2026-09-01 before the rewrite, and it is not salvageable. Its only output call is `Scene2D::append_vello_scene`, deleted from the contract by 937674176, so it cannot compile; built around a `vello::Scene`, it could never have run on the hybrid engine. Beyond that: every character is laid out as its own parley layout so there is no kerning between adjacent glyphs, `layout_row` concatenates advance widths so there is no math spacing at all, `\text{}` discards its own spaces, `\left\langle` renders the literal letters `langle`, 11 of the 24 lowercase Greek letters are missing and an unknown command panics inside `View::body`, colour falls back to a hardcoded white, and the radical is positioned by sampling 96 points per Bézier segment against tuned constants — which the replacement does in a few lines of arithmetic over the `MATH` constants the font already supplies.

The full audit is in #213's "prior art" section. The one part worth reading is `build_radical_outline_from_assembly` / `build_radical_outline_from_math`, which handles `MathVariants` correctly and is the single architectural call this implementation got right; the replacement generalises it from a radical special case to the mechanism every stretchy glyph uses.

The `rescue/waterui-math` branch is untouched, so nothing is lost by closing this.

PR #159, opened 2026-08-29
fix(release): remove extracted filtrate overrides from release-plz.toml

lexoliu:
Removes  and  overrides from  so the release-plz action on  can proceed with publishing.


lexoliu:
Closing this pre-release PR; further pre-release work will land on dev first.

PR #303, opened 2026-09-03
refactor(cef): retire the CEF-specific input entry points for the GpuSurface carrier

### LINKED ISSUE #252: Retire the CEF-specific input entry points in favour of the GpuSurface input carrier
## Problem

PR #250 (issue #197) gives `GpuSurface` a backend-neutral input carrier over the C ABI: `waterui_gpu_surface_send_input_event`, `waterui_gpu_surface_wants_input_events`, `waterui_gpu_surface_ime_caret`, plus the generic Apple responder (`WuiGpuSurfaceInput.swift`) and Android sink (`GpuSurfaceInput.kt`).

`ffi/src/components/platform/browser_cef.rs` still exposes thirteen CEF-specific input/focus/scroll/edit entry points, and `CefSurfaceView.swift` in the Apple backend drives the engine through them, even though `components/platform/browser-cef` already implements the generic `GpuView` contract (`CefInputGpuView` in `src/input.rs`). Two parallel input paths for the same surface is exactly the duplication the generic carrier exists to remove.

## Change

- Switch `CefSurfaceView.swift` onto the generic `WuiGpuSurfaceInput` responder introduced by #250 and delete its engine-specific key/scroll/IME plumbing.
- Delete the thirteen CEF-specific input/focus/scroll/edit entry points from `ffi/src/components/platform/browser_cef.rs` and regenerate `ffi/waterui.h`. `waterui_cef_prepare_macos_application` and `waterui_cef_run_packaged_subprocess` stay.
- Acceptance is the macOS real-engine check run for the CEF browser (typing, scrolling, IME composition through the generic path).

Both halves land together (monorepo FFI + `backends/apple` submodule pin), since the Swift side cannot build against a header that has lost the old symbols.

lexoliu:
Fixes #252. Stacked on #302 (its commit is included here until it merges).

- `ffi/src/components/platform/browser_cef.rs`: the surface is built from `gpu_view_with_input`, so it reports `wants_input_events` and the host's generic responder delivers pointer, keyboard, scroll and composition events through `waterui_gpu_surface_send_input_event`. The fourteen CEF-specific entry points (focus, pointer move/button, scroll, key, commit text, set/finish/cancel composition, edit, go back/forward, set viewport, request frame) and the `WuiCefInputModifiers` / `WuiCefPointerButton` / `WuiCefEditCommand` types are gone. `waterui_force_as_cef_webview`, `waterui_force_as_chromium`, `waterui_cef_surface_drop`, `waterui_cef_prepare_macos_application` and `waterui_cef_run_packaged_subprocess` stay.
- `components/platform/browser-cef/src/gpu/mod.rs`: `request_browser_frame` requests the next redraw on macOS as it already did elsewhere. The browser runs with external begin frames, so it composites only when asked; the Apple display link that paced those requests lived in the retired Swift plumbing.
- `backends/apple` (water-rs/apple-backend `agent/cef-generic-input`): `CefSurfaceView` reduced to a host that fills its box and keeps the semantic view alive. `backends/android`: header only.
- Header regenerated (all three copies).

Verification on macOS (`water run --platform macos --path examples/webview-cef`): the page renders through the generic surface and the frame loop runs continuously (about 328 renders in a few seconds, CEF IOSurface imports logged), which also exercised #302. **Typing, scrolling and IME composition through the generic path are not yet verified on screen**: this machine's terminal has no Accessibility permission, so neither CGEvent posting nor AppleScript can reach the window, and attaching a debugger needs Developer Mode. Draft until that check has run.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -25,9 +25,9 @@ waterui-preview-version = "0.1.0"
 waterui-preview-protocol-version = "0.1.0"
 android-kotlin-version = "2.3.21"
 apple-backend-url = "https://github.com/water-rs/apple-backend.git"
-apple-backend-commit = "9799defb87f4c3b15dd8e26f44ac2381adbeee18"
+apple-backend-commit = "3b36eb3932d6ad5cec8a7e31a89f599dfa3edd28"
 android-backend-url = "https://github.com/water-rs/android-backend.git"
-android-backend-commit = "f4649a3641f0442bcbbfd1de2994af111d234c2f"
+android-backend-commit = "0e04516d1b6c0194997760aa76e2a98a3ba11456"
 
 [[bin]]
 name = "water"
--- a/components/platform/browser-cef/src/gpu/mod.rs
+++ b/components/platform/browser-cef/src/gpu/mod.rs
@@ -42,12 +42,16 @@ fn sync_browser_viewport(page: &CefPageHandle, frame: &GpuFrame<'_>) -> f64 {
     scale
 }
 
-#[cfg(target_os = "macos")]
-fn request_browser_frame(page: &CefPageHandle, _frame: &mut GpuFrame<'_>) {
-    page.request_frame();
-}
-
-#[cfg(not(target_os = "macos"))]
+/// Asks Chromium for a frame, and asks the surface for the next one.
+///
+/// The browser is created with external begin frames, so it composites only
+/// when told to: this call is its clock, and the surface's own redraw loop is
+/// what ticks it. Every render therefore requests the next, and the loop runs at
+/// the display's rate for as long as the surface is attached — on every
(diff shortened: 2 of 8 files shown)


lexoliu:
Superseded by #404, which is open, green, and carries the acceptance evidence. This draft was cut on 2026-09-03, before the generic input carrier (#250) landed on `dev`, so it also contains that plumbing (`ffi/src/components/visual/gpu_surface.rs`, `applied_filter.rs`); all of that is on `dev` now. #404 is the remaining half: the CEF surface is built with `gpu_view_with_input`, the eleven redundant `waterui_cef_surface_*` input exports are gone, and `CefSurfaceView` hosts the generic responder. It also fixes a re-entrancy panic this branch would have hit (a redraw wake arriving inside the Rust render call, which re-measured through a live `RefCell` borrow).

Issue #312, opened 2026-09-03
water run: the generated FFI crate resolves its own Cargo.lock instead of the project's, so a fresh registry publish can break the app build


lexoliu:
## Problem

`water run` builds an application through the generated FFI crate under `~/.water/build_cache/<project>/managed_backends/ffi/`. That crate has its own `Cargo.lock`, resolved fresh from the registry the first time it is generated, so the dependency set the app is built with is not the one the project (or, for workspace members, the workspace) pins and tests.

Observed today on `examples/map`: the workspace `Cargo.lock` pins `tinyvec 1.12.0`, `examples/image`'s managed crate (generated earlier) also has 1.12.0, but `examples/map`'s managed crate resolved `tinyvec 1.13.0` — published since — which does not compile on stable 1.98:

```
error: cannot find macro `vec` in this scope
 --> tinyvec-1.13.0/src/tinyvec.rs
note: `vec` is imported here, but it is a module, not a macro
  3 | use alloc::vec::{self, Vec};
```

so `water run --platform macos` fails in a dependency that CI never sees, while every workspace check is green. Any freshly published dependency with a regression can break `water run` this way, and two examples in the same workspace build different dependency versions.

## Expected

The managed FFI crate resolves against the project's lockfile: when the project is a workspace member, the workspace `Cargo.lock`; otherwise the project's own `Cargo.lock`. Regenerating the managed crate must not re-resolve versions the project already pins. (`cargo`'s `--locked`/copying the lockfile into the generated crate before the first resolve are the obvious mechanisms; `cargo` will keep the pinned versions and only add entries for crates the lockfile does not know.)

Found while verifying #173. The tinyvec 1.13.0 upstream regression is separate from this lockfile-isolation defect.

lexoliu:
Same failure on the Hydrolysis path: `water run --platform macos --backend hydrolysis` for `examples/map` builds its own managed hydrolysis crate, which also resolved `tinyvec 1.13.0` and fails to compile, while the workspace lock pins 1.12.0. Both managed crates (ffi and hydrolysis) need to resolve against the project's lockfile.

PR #315, opened 2026-09-03
fix(map): let the Apple backend declare its MapKit realization so the GPU map yields at runtime

### LINKED ISSUE #173: Runtime backend discrimination for map components (macOS Hydrolysis vs MapKit)
## Problem

In `examples/map`, `waterui_map_gpu::install` is currently gated with `#[cfg(not(target_vendor = "apple"))]`.

This was necessary because on Apple targets, the native MapKit bridge intercepts `Map` by type ID before body evaluation. If `map_gpu` installs unconditionally, it intercepts via `Hook<MapConfig>` on both macOS Hydrolysis and native AppKit.

Consequently, running `examples/map` on macOS under the Hydrolysis renderer currently renders Hydrolysis's placeholder map widget rather than GPU vector tiles.

## Proposed Solution

- Implement a runtime backend capability / realization detection mechanism.
- Allow macOS applications to dynamically select GPU vector tile map rendering when running under Hydrolysis, while preserving native MapKit realization when running under the Apple native backend.

lexoliu:
Fixes #173

`examples/map` gated `waterui_map_gpu::install` behind `cfg(not(target_vendor = "apple"))` because nothing told the environment that the Apple backend brings a native map, so on macOS under Hydrolysis the example rendered the placeholder widget instead of GPU vector tiles.

- The FFI init the native backends call (`waterui_init`) now runs `__configure_native_realizations`, which declares the realizations that backend brings: on Apple with the `map` feature it inserts the `Hook<MapConfig>` that `waterui_map_gpu::install` already documents as its yield signal — the hook renders the configuration into the raw view unchanged, so MapKit is reached through `waterui_force_as_map` exactly as before. The CEF runtime installation moves under the same function. Hydrolysis never runs that init, so there the application's GPU map installs.
- `examples/map` installs the GPU map unconditionally.

Verified on macOS, both realizations, by screenshot: `water run --platform macos` shows Apple Maps (MapKit) with the Manhattan overlay; `water run --platform macos --backend hydrolysis` shows the OpenFreeMap positron vector tiles with the Rust-drawn controls. Clippy on `waterui-ffi` with `--features map` and `--features map,cef-header` is clean, and `map-example` checks.

Note: both runs needed the managed crates' lockfiles pinned back to the workspace's `tinyvec 1.12.0` by hand — see #312.

### DIFF
--- a/examples/map/src/lib.rs
+++ b/examples/map/src/lib.rs
@@ -181,7 +181,6 @@ pub fn app(mut env: Environment) -> App {
     env.insert(MapGpuOptions::new(Url::new(
         "https://tiles.openfreemap.org/styles/positron",
     )));
-    #[cfg(not(target_vendor = "apple"))]
     waterui_map_gpu::install(&mut env);
     let state = MapExampleState::new();
     App::new(move || content(state.clone()), env)
--- a/ffi/src/components/data/map.rs
+++ b/ffi/src/components/data/map.rs
@@ -6,8 +6,24 @@
 use crate::reactive::{WuiBinding, WuiComputed};
 use crate::{IntoFFI, WuiStr};
 use alloc::vec::Vec;
+#[cfg(target_vendor = "apple")]
+use waterui_core::{Environment, view::ViewConfiguration as _};
 use waterui_map::{Annotation, Coordinate, Location, MapConfig, MapStatus, MapStyle, Region};
 use waterui_str::Str;
+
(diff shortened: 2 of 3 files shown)


lexoliu:
Merging with two red legs that predate this PR and are tracked separately: Hygiene (#316, fixed by #317) and Bench (windows-latest) (#320, the stress bench timing out on runners without a Vulkan ICD). Every other check is green.

Issue #316, opened 2026-09-03
CI Hygiene: rustsec/audit-check compiles cargo-audit from source and now fails on tinyvec 1.13.0


lexoliu:
The Hygiene job's "Security audit" step (`rustsec/audit-check@v2`) fails on every PR since today (first seen on #315, run job "Hygiene"):

```
error: cannot find macro `vec` in this scope
error: could not compile `tinyvec` (lib) due to 1 previous error
error: failed to compile `cargo-audit v0.22.2`
```

`audit-check` builds `cargo-audit` from source with a fresh resolution on every run, so a freshly published broken dependency (tinyvec 1.13.0, which does not compile on stable 1.98 — the same regression as #312) breaks our gate without any change on our side. The rest of the Hygiene job already uses `taiki-e/install-action` for `cargo-deny`, which downloads a prebuilt binary.

Fix: install `cargo-audit` the same way (`taiki-e/install-action@cargo-audit`) and run `cargo audit` directly. Coverage is unchanged — same tool, same advisory database — the tool's own build just stops being part of our critical path.

PR #318, opened 2026-09-03
chore: refresh Cargo.lock for the kit pin on main

### LINKED ISSUE #305: Cargo.lock is stale after the kit pin to 7d8b386
PR #293 moved `kit` to 7d8b386, where `waterkit-fs` no longer depends on `js-sys` and `wasm-bindgen`, but `Cargo.lock` still lists them under `waterkit-fs`. Every plain build rewrites the lockfile (two deletions), which shows up as an unrelated `Cargo.lock` change in whatever branch happened to build, and a `--locked` build of dev would fail. Refresh the lockfile once.

lexoliu:
Cherry-pick of 79f2c8858 (#306, Fixes #305 on `dev`) onto `main`, the same release-chain pattern as #286 and #300.

Release run 33807655083 publishes every crate up to `waterui` itself and then fails: `cargo publish` refuses because `Cargo.lock` is dirty — cargo rewrites it on first use because `main`'s lock still lists `js-sys`/`wasm-bindgen` under `waterkit-fs` after the kit pin moved (#305). Subdirectory crates publish regardless because the root lock is not part of their package; the root `waterui` crate is where it bites. With the lock committed, the rerun publishes `waterui` and the CLI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11950,13 +11950,11 @@ dependencies = [
  "dirs",
  "indexed_db_futures",
  "jni 0.22.4",
- "js-sys",
  "serde",
  "serde_json",
  "swift-bridge",
  "thiserror 2.0.20",
  "tracing",
- "wasm-bindgen",
  "waterkit-build",
  "windows 0.62.2",
 ]


lexoliu:
Hygiene is red here for the reason this branch predates: `rustsec/audit-check` compiling cargo-audit against the broken tinyvec 1.13.0 (#316). #317 fixed that on `dev`; `main` picks it up with the next dev→main promotion. Every other check is green, and this PR is lockfile-only.

Issue #321, opened 2026-09-04
Release stops at hydrolysis-m3: versioned dev-dependencies on unpublished workspace crates cannot be packaged


lexoliu:
## Symptom

The Release run for the post-#318 `main` commit publishes 27 packages, including `waterui 0.3.0` and `waterui-backend-core 0.1.0`, and then stops:

```
ERROR failed to release package
    failed to publish hydrolysis-m3: Packaging hydrolysis-m3 v0.1.0
    error: failed to prepare local package for uploading
    Caused by:
      no matching package named `waterui-canvas` found
      location searched: crates.io index
      required by package `hydrolysis-m3 v0.1.0 (backends/hydrolysis_m3)`
```

## Cause

`backends/hydrolysis_m3/Cargo.toml` has `waterui-canvas.workspace = true` under `[dev-dependencies]`, and the workspace entry carries `version = "0.1.0"`. A dev-dependency with a version requirement stays in the packaged manifest, so `cargo package` insists on resolving it from the registry, and `waterui-canvas` has never been published. release-plz orders publishes by normal dependencies only, so nothing makes it publish canvas first.

Four more publishable crates inherit the same versioned workspace entry for a crate that is not on the registry:

| crate | dev-dependency | on crates.io |
| --- | --- | --- |
| `hydrolysis-m3` | `waterui-canvas` | no |
| `hydrolysis` | `waterui-canvas` | no |
| `waterui-dew` | `waterui-canvas`, `waterui-chart` | no |
| `waterui-testing` | `waterui-canvas` | no |
| `waterui-text` | `hydrolysis-m3` | no |

`waterui-text` is not merely blocked: its dev-dependency on `hydrolysis-m3`, which depends on `waterui` and so back on `waterui-text`, closes a cycle. No publish order can satisfy it while the version requirement is there. The 36 example crates carry the same inherited dev-dependencies but all set `publish = false`, so they are unaffected.

## Fix

Declare these dev-dependencies as plain path dependencies, which is what the workspace already does for `waterui-testing`. Cargo drops a path-only dev-dependency when packaging, the tests still build from the path, and no publish order has to satisfy them.

PR #323, opened 2026-09-04
chore(release): make workspace-internal dev-dependencies path-only (main)

lexoliu:
Same fix as #322, applied to `main` where the Release workflow runs. Diagnosis in #321.

The current Release run published 27 packages including `waterui 0.3.0` and `waterui-backend-core 0.1.0`, then stopped at `hydrolysis-m3`: its test-only dependency on `waterui-canvas` carries a version requirement, `waterui-canvas` has never been published, and release-plz orders publishes by normal dependencies so nothing publishes canvas first. Four crates on `main` inherit that versioned entry (`hydrolysis-m3`, `hydrolysis`, `waterui-dew`, `waterui-testing`); each now uses a plain path dependency, as the workspace already does for `waterui-testing` elsewhere. Every rerun stops in the same place until `main` carries this.

**Left unmerged for you.** Merging into `main` re-triggers publishing, and that is your decision. The crates that would then publish are the same set the current run is already working through; this only removes the obstacle.

### DIFF
--- a/backends/dew/Cargo.toml
+++ b/backends/dew/Cargo.toml
@@ -100,7 +100,7 @@ waterui-testing = { path = "../../testing" }
 # `waterui-graphics` for its `gpu` feature (`CanvasImage` decodes through it),
 # which is why they are confined to dev-dependencies — the firmware graph,
 # built from the library alone, stays GPU-free.
-waterui-canvas.workspace = true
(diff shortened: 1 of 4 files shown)


lexoliu:
All checks green and the commit is signed. Ready whenever you want to merge; that is what restarts publishing.

PR #334, opened 2026-09-05
chore(svg): consume waterui-svg 0.1.0 from crates.io and drop the in-tree copy

### LINKED ISSUE #186: Transition waterui-svg to standalone water-rs/svg repository
## Description

`water-rs/svg` is established with standalone CI workflows.

## Tasks

- Once published to crates.io, update monorepo `Cargo.toml` to consume `waterui-svg` from registry.
- Remove `components/visual/svg/` directory from monorepo.
- Verify workspace builds and tests pass cleanly.

## Standalone repository side

- After the crate is published, replace the standalone repository's `git = "https://github.com/water-rs/waterui.git", branch = "dev"` dependencies on `waterui-core`, `waterui-graphics`, `waterui`, and `waterui-testing` with registry version pins, so the repository stops tracking the monorepo's integration branch and builds against released APIs.
- Landing the in-tree removal requires `wgpu-external-frame` to be consumed from crates.io first (#176); otherwise the monorepo release would publish a second copy of a crate the standalone repositories already own.

lexoliu:
Stacked on #333 (retarget to `dev` once it merges). Same shape as the canvas flip: `waterui-svg = "0.1.0"` from crates.io (published tonight from this tree), `components/visual/svg` deleted, and `[patch.crates-io]` extended with `waterui-layout`, `waterui-macros` and `waterui-str`, which the registry crate depends on; the lock holds one copy of each (verified).

Verified locally with `cargo check --tests` for waterui, dew, icon and fontawesome7 (with and without their `svg` feature). The standalone `water-rs/svg` pins follow in that repository.

Fixes #186


lexoliu:
Parked as draft — this must not merge as is. The in-tree svg has two commits after `waterui-svg 0.1.0` was published (390a9c368 `feat(graphics)!: give SceneContent an intrinsic size`, d2aba8093 `fix(svg): cap an SVG at its intrinsic size`), so the registry crate compiles against the patched tree (the trait method has a default) but silently loses the intrinsic-size behaviour. The flip needs `waterui-svg 0.1.1` on crates.io first, and that in turn needs the monorepo to publish the `waterui-graphics` release carrying the breaking `intrinsic_size` change (0.4.0), because the standalone `water-rs/svg` dev already targets it. Order: graphics 0.4.0 → svg 0.1.1 → this PR (rebased). #335 and #336 are being restacked directly on #333 since chart and map-gpu do not depend on svg.

Issue #343, opened 2026-09-05
Apple: WuiMapView no longer compiles under WATERUI_MAP since apple-backend#11 removed makeRegionWatcher; CI never builds that shape


lexoliu:
`water run --platform macos` on `examples/map` fails in the Swift package:

```
Sources/WaterUI/Components/Media/WuiMapView.swift:137:21: error: cannot find 'makeRegionWatcher' in scope
```

Cause: water-rs/apple-backend#11 (the periphery cleanup, 92389f0) deleted `makeRegionWatcher` from `Sources/WaterUI/Reactive/Watcher.swift` because periphery reported it unused. Its only caller sits inside `#if WATERUI_MAP` in `WuiMapView.swift`, and neither the apple CI builds (`swift build … -warnings-as-errors`) nor the periphery scan pass `-DWATERUI_MAP`, so the map shape was invisible to both: periphery saw dead code, CI saw a clean build. The CLI does compile it (`apple_swift_conditions` in `cli/src/apple/platform.rs` adds `-DWATERUI_MAP` when the app links the map capability and `-DWATERUI_NO_GPU` when it drops the GPU one), so the break only shows up in a real app build.

The same gap covers the `WATERUI_NO_GPU` shape (15 `#if !WATERUI_NO_GPU` sites): nothing in CI compiles it either.

Fix, in the apple-backend repo: restore `makeRegionWatcher`, and make CI and the periphery scan compile every Swift condition the CLI can emit (`-DWATERUI_MAP`, `-DWATERUI_NO_GPU`) so a conditional-only symbol is neither "unused" nor unbuilt.

Regression introduced by my own cleanup; found while verifying #312 on `examples/map`.

PR #344, opened 2026-09-05
fix(cli): seed the managed FFI crate's Cargo.lock from the project lockfile

### LINKED ISSUE #312: water run: the generated FFI crate resolves its own Cargo.lock instead of the project's, so a fresh registry publish can break the app build

lexoliu:
The managed FFI crate under `~/.water/build_cache/…/managed_backends/ffi` is its own Cargo workspace and resolved its graph fresh from the registry, so `water run` built the app with versions nothing in the project pins or tests; `examples/map` picked up `tinyvec 1.13.0` that way and failed to compile while every workspace check was green (#312).

- `Project` resolves the workspace root alongside the target directory from the single `cargo metadata --no-deps` run it already makes (`CargoLayout`), and exposes `lockfile_path()`: the workspace `Cargo.lock` for a member, the project's own otherwise.
- `templates::ffi::seed_lockfile` copies that lockfile into the managed crate before Cargo resolves. Cargo keeps every pinned version and only adds the FFI-only entries on top. A copy of the seed (`Cargo.lock.seed`) sits beside the managed lockfile so re-seeding happens only when the project's pins change or the managed lockfile is missing — never on every scaffold, which would re-resolve the FFI-only entries (and hit the index) each run. A project with no lockfile yet is left to resolve on its own.
- Unit test covers first seed, Cargo's rewrite being left alone, re-seed on changed pins, and re-seed after deletion.

Verified end to end on `examples/map` with the managed `Cargo.lock` deleted: the CLI wrote the 353 KB workspace lockfile in as the seed, Cargo pruned it to the 211 KB the FFI graph needs, and the managed lockfile carries `tinyvec 1.12.0` (the workspace pin) instead of a fresh resolution.

Fixes #312

### DIFF
--- a/cli/src/project_model/project.rs
+++ b/cli/src/project_model/project.rs
@@ -14,12 +14,24 @@ enum OpenMode {
     PreviewBuild,
 }
 
-fn spawn_target_dir_resolution(
+/// What `cargo metadata` reports about the tree a project builds in.
+///
+/// Resolved once per [`Project`] and shared by everything that needs it, so
+/// one `cargo metadata` run serves the target directory and the lockfile.
+#[derive(Debug, Clone)]
+struct CargoLayout {
+    target_dir: PathBuf,
+    /// Root of the Cargo workspace the project belongs to — the project itself
+    /// when it is not a workspace member. This is where its `Cargo.lock` lives.
+    workspace_root: PathBuf,
+}
+
+fn spawn_cargo_layout_resolution(
     current_dir: &Path,
-) -> Shared<BoxFuture<'static, Result<PathBuf, String>>> {
+) -> Shared<BoxFuture<'static, Result<CargoLayout, String>>> {
     let current_dir = current_dir.to_path_buf();
     smol::spawn(async move {
-        get_target_dir(&current_dir)
+        resolve_cargo_layout(&current_dir)
             .await
             .map_err(|error| error.to_string())
     })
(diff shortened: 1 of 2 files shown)


lexoliu:
End-to-end run completed: with the apple-backend fix from water-rs/apple-backend#13 checked out, `water run --platform macos` on `examples/map` seeded the managed lockfile, built, and reached `Application started` (Metal adapter selected, GPU surface attached). The first two attempts failed in the Swift package for the unrelated #343 regression, which is what surfaced that issue.

Issue #320, opened 2026-09-03
Windows bench leg: stress steady_redraw times out after 40 minutes on runners without a Vulkan ICD


lexoliu:
On PR #315's CI run 33809384571 the `Bench (windows-latest)` leg (step "Bench examples/stress") timed out:

```
SLOW [>300.000s] stress-example bench::waterui_bench_stress_steady_redraw
…
TIMEOUT [2404.601s] (1/1) stress-example bench::waterui_bench_stress_steady_redraw
Summary [2405.297s] 1 test run: 0 passed, 1 timed out
bench run failed: a bench panicked or exceeded its budget
```

The runner has no Vulkan ICD (`windows_read_data_files_in_registry: Registry lookup failed to get ICD manifest files. Possibly missing Vulkan driver?`), so wgpu is on whatever adapter it falls back to, and the steady-redraw bench never reaches its sample count within nextest's 40-minute slow-test ceiling. That is 40 minutes of Windows runner time per selected run for a red leg that says nothing about the change under test.

The Ubuntu leg of the same bench passes. Related: #152 (Windows leg resurrected; slowness was left open), #264 (bench path filter, which is why only PRs touching bench-relevant paths see this). Fix should be one of: make the stress bench pick a real adapter or skip when only a non-hardware adapter exists on Windows, or bound the bench by wall-clock so it fails fast with a measurement instead of timing out.

Pre-existing; found on #315, which touches ffi/ and examples/map only.

lexoliu:
Measured on run 33945713983 (draft #351, Windows shape reduced to `--warmups 1 --samples 4 --repetitions 1` so the leg completes), Linux at the normal `2/24/3` shape for comparison. `stress_steady_redraw`, per frame:

| | Windows (WARP, no Vulkan ICD) | Linux (llvmpipe) |
|---|---|---|
| frame mean | 6 983 ms | 337 ms |
| `scene_dispatch` | 6 933 ms | 327 ms |
| `render` | 50 ms | 9 ms |
| applied filters per frame | 144 | 144 |
| `applied_filter_capture_us` per frame | ≈ 4 500 ms | ≈ 260 ms |
| per filter capture | ≈ 39 ms | ≈ 1.8 ms |

Wall clock on the leg: 23 min cold CLI build, 22 min for four samples of the stress bench, 1 min for the list bench, 8 min cache save.

[...]

PR #351, opened 2026-09-05
ci(bench): measurement run for #320 (do not merge)

lexoliu:
Measurement only, for #320: the Windows stress bench times out before printing its report, so this run uses a five-frame shape on Windows to get the per-phase breakdown (dispatch vs render) behind the ~30 s/frame. Closed once the numbers are on the issue.


lexoliu:
Measurement recorded on #320; the reduced shape was never meant to merge.

PR #387, opened 2026-09-07
fix(hydrolysis): auto-fallback to software adapter when no GPU is available

### LINKED ISSUE #386: Hydrolysis should auto-fallback to software adapter when no GPU is available
## Summary
On a machine without a discrete/integrated GPU, Hydrolysis panics even though a perfectly usable compute-capable software Vulkan adapter (e.g. Mesa `llvmpipe`) is available. It only succeeds if `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` is set manually.

## Expected behavior
When no hardware GPU adapter is found, Hydrolysis should automatically fall back to a compute-capable software adapter. The `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` environment variable should only be required to *force* the use of a software adapter when a hardware GPU is also present.

## Actual behavior
`request_hydrolysis_adapter` in `backends/hydrolysis/src/platform.rs` uses `AdapterSelection::PRODUCTION` for normal (non-test) surfaces, which sets `allow_software_adapter: false`. It enumerates all adapters, skips any `Cpu`/`Noop` adapter unless `allow_software_adapter` is true, and then panics with:

```
failed to find a compute-capable modern adapter. Surface-compatible adapters inspected: 'llvmpipe (LLVM 15.0.7, 256 bits)' (Gl, Cpu, compute=true, max_compute_workgroups_per_dimension=65535). Set WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1 to explicitly allow software fallback adapters for diagnostics.
```

## Steps to reproduce
1. Start with a fresh Ubuntu 22.04 VM with no GPU (or inside Xvfb).
2. Install `mesa-vulkan-drivers` so `llvmpipe` is the only Vulkan ICD.
3. Run the Hydrolysis example without the fallback env var:
   ```bash
   cargo run -p hydrolysis --example wayland_showcase --features winit
   ```
4. Observe the panic above.
5. Re-run with `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` and it works.

## Suggested fix
Track a "fallback candidate" while enumerating adapters. If a better (non-CPU) adapter is found, use it. If the loop completes without a better candidate, automatically use the best compute-capable software adapter instead of panicking. Keep the existing env var as an override to force software fallback even when a hardware adapter is available.

## Environment
- OS: Ubuntu 22.04
- Rust: 1.98.1
- WaterUI: dev branch (post 0.3.0 Pre-1.0)
- GPU: none (Mesa lavapipe / llvmpipe as CPU Vulkan ICD)

lexoliu:
## Summary

On machines without a hardware GPU, Hydrolysis would panic with `failed to find a compute-capable modern adapter` even when a perfectly functional software Vulkan adapter such as Mesa `llvmpipe` was available. Users had to discover and set `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` before the backend would accept a CPU adapter.

This change makes the adapter selection in `request_hydrolysis_adapter` automatically reserve a compute-capable software adapter as a fallback candidate while still preferring any hardware adapter it finds. If no non-software adapter is available, the fallback is selected automatically, so apps run out of the box on headless CI runners and GPU-less VMs. The existing `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` override continues to force software fallback even when a GPU is present.

High-level change in `backends/hydrolysis/src/platform.rs`:

```rust
// Before
if info.backend == wgpu::Backend::Noop
    || (info.device_type == wgpu::DeviceType::Cpu && !selection.allow_software_adapter())
{
    continue;
}

// After
try { best_hardware } else { best_software_fallback }
```

`fallback_candidate` is tracked separately from `best_candidate` and only used when no hardware adapter is found. The panic messages now also tell users to use `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` for diagnostics, rather than requiring it for normal operation.

Closes #386.

### DIFF
--- a/backends/hydrolysis/src/platform.rs
+++ b/backends/hydrolysis/src/platform.rs
@@ -654,6 +654,7 @@ async fn request_hydrolysis_adapter(
 
         let backends = wgpu::Backends::from_env().unwrap_or(wgpu::Backends::all());
         let mut best_candidate: Option<(AdapterPreference, wgpu::Adapter)> = None;
+        let mut fallback_candidate: Option<(AdapterPreference, wgpu::Adapter)> = None;
         let mut inspected_adapters: Vec<String> = Vec::new();
 
         for adapter in instance.enumerate_adapters(backends).await {
@@ -687,20 +688,32 @@ async fn request_hydrolysis_adapter(
                 limits.max_compute_workgroups_per_dimension
             ));
 
-            if info.backend == wgpu::Backend::Noop
-                || (info.device_type == wgpu::DeviceType::Cpu
-                    && !selection.allow_software_adapter())
-            {
+            if info.backend == wgpu::Backend::Noop {
                 tracing::info!(
                     target: "hydrolysis::gpu",
                     context,
                     adapter = ?info,
-                    "skipping software/noop adapter because fallback adapter was not requested"
+                    "skipping noop adapter"
                 );
                 continue;
             }
 
             if !compute_capable {
+                continue;
+            }
+
+            if info.device_type == wgpu::DeviceType::Cpu && !selection.allow_software_adapter() {
+                let preference = AdapterPreference::for_info(&info);
+                match &fallback_candidate {
+                    Some((best_preference, _)) if *best_preference <= preference => {}
+                    _ => fallback_candidate = Some((preference, adapter)),
+                }
+                tracing::info!(
+                    target: "hydrolysis::gpu",
+                    context,
+                    adapter = ?info,
+                    "software adapter reserved as automatic fallback"
+                );
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lexoliu:
Closing with #386: this is the documented design, not a defect. AGENTS.md ("Rendering backend philosophy: Hydrolysis vs Dew") states that Hydrolysis is GPU-first and GPU-required, that there is no CPU rasterization path, and that a CPU fallback must never be added; Principle 4 forbids switching realisation at runtime when the intended path is unavailable. A machine with only `llvmpipe` is what the panic exists for — silently running a renderer built for full-scene redraw at 120fps on a software Vulkan ICD produces an app at a few frames per second and hides the real problem, that the deployment has no GPU. `WATER_HYDROLYSIS_FORCE_FALLBACK_ADAPTER=1` is the deliberate diagnostic override. For a GPU-less target the framework's answer is the Dew backend (CPU-first, dirty-rect) or GTK on Linux desktops.

PR #393, opened 2026-09-07
fix(hydrolysis): rasterize layers with multisampling so a scene renders to the same bytes twice

### LINKED ISSUE #271: hydrolysis: text rasterization differs by one antialiased pixel between rebuilds of identical source
Rendering `waterui_tree_grid.png` from two builds of the same source (found while verifying #270 with byte comparison of all 19 dew/hydrolysis exports) produced a one-pixel difference at (200, 133), inside a glyph, of 1/255 per channel. Nothing gates on it today (no baseline PNGs are committed), but it puts a floor under any future pixel-exact snapshot gate for this backend. Find the non-determinism (parley/skrifa glyph cache order, hash-map iteration in the text pre-pass, float summation order across threads in the parallel scene build) and remove it so identical input yields identical bytes; add a determinism test that renders the same text scene twice in one process and byte-compares.

lexoliu:
Rendering an unchanged scene through vello's analytic `Area` antialiasing does not produce the same pixels twice.

## Root cause

`path_count.wgsl:189` assigns each path segment its slot inside a tile with an `atomicAdd`, so the order the `fine` stage walks a tile's segment list in is whatever order that dispatch happened to claim the slots. `fine.wgsl`'s `fill_path` then integrates coverage as `area[i] += a * dy` over that list — a floating-point sum, and floating-point addition is not associative. Two renders of a byte-identical scene therefore land an ULP apart on antialiased edges, which quantizes to pixels 1/255 apart.

Nothing on the WaterUI side was at fault: the vello `Scene` encoding handed to `render_to_texture` hashes identically on every frame, and the divergence appears in the layer texture that comes back.

## Evidence

Bisected by hashing the scene encoding and reading back the vello layer texture before compositing, on the tree-grid screen (440x920, one Vello layer):

| observation | `Area` (before) | `Msaa16` (after) |
| --- | --- | --- |
| vello scene encoding hash, 12 pumps of one runtime | identical every pump | identical every pump |
| vello layer texture, consecutive pumps of that identical scene | 454-794 differing bytes each pump, all +/-1..3 on glyph edges | 0 differing bytes, all 12 pumps |
| composited window frame, 8 fresh runtimes in one process | differs at (200, 133) g: 139 vs 140 | byte-identical |
| determinism test, 6 separate processes | 5 of 6 FAIL, always (200, 133) g, flipping 139 <-> 140 | 6 of 6 pass |
| `render_tree_grid_snapshot` PNG sha, 6 separate debug processes | all 6 equal (a single frame lands on the same side most of the time) | all 6 equal |
| same export, 2 release processes | the two differ from each other; run 1 matches debug, run 2 does not | both equal, and equal to debug |

(200, 133) is the pixel #271 reported, inside the `Contour` button's label.

Frame instant is not involved: pumping one runtime twice at the *same* `Instant` reproduces the difference, and pumping at +500 ms does not change it. Neither is `encode_vello_layers_parallel` — this scene has a single Vello layer, so the parallel branch is never entered.

## The change

The multisampled fine stage accumulates winding numbers and sample masks with integer `atomicAdd`s (`fine.wgsl` msaa section), and integer addition does not care what order it is applied in, so the same scene always composites to the same bytes. Every Hydrolysis layer now renders with `Msaa16`, expressed as one `LAYER_ANTIALIASING` constant with the matching single-variant `AaSupport` derived from it, so a renderer still compiles exactly one fine-stage variant.

## Trade-off

The two rasterizers do not agree pixel-for-pixel. On the tree-grid frame 0.4% of bytes move, by 6.4/255 on average and 27/255 at most; on the MD3 shape set 2.6% of bytes move, by at most 10/255. Both pairs are indistinguishable side by side at 3x — text stems and curved shape edges look the same.

No cost was measurable on Apple Silicon: 120 captured frames of 160 text buttons at 1600x1200, median 9.59 ms under `Area` against 9.39 ms under `Msaa16` (p10 9.00 vs 8.84). vello's own docs say `Area` "generally has better performance than the multi-sampling methods", so this deserves a look on a bench lane and on a mobile GPU before it is taken as free everywhere. `Msaa16`'s workgroup storage (~4.6 KB) is inside the WebGPU baseline of 16 KB, so it is not expected to fail pipeline creation on any adapter that already runs Hydrolysis.

Reverting is three lines: set `LAYER_ANTIALIASING` back to `vello::AaConfig::Area`.

## Test

`tree_grid_frame_is_byte_identical_across_runtimes` renders the tree-grid screen through eight independently built runtimes — fresh view tree, renderer and retained scene each, sharing only the wgpu device — and requires all eight RGBA buffers to be byte-identical, printing the first differing pixel, channel, both values and the differing-byte count when they are not. Each run is pumped at its own `Instant::now()`, since the screen declares no animation and must not depend on the wall-clock phase it was captured at. The screen builder is now shared with `render_tree_grid_snapshot` instead of duplicated.

The test fails on the parent commit.

## Verification

- `cargo clippy -p hydrolysis --all-targets -- -D warnings`: clean
- `cargo nextest run -p hydrolysis`: 210 passed, 1 skipped (209 passed / 1 failed before the fix, the failure being the new test)
- `cargo nextest run -p hydrolysis-m3`: 191 passed, 24 skipped
- MD3 shape-set and tree-grid PNGs reviewed by eye under both rasterizers

## Left as-is, deliberately

`components/visual/graphics` still renders Scene2D GPU surfaces with `AaConfig::Area` (`gpu/shared_context.rs:193`, `scene/scene_surface.rs:268`), so a scene drawn through a `GpuSurface` keeps the same non-determinism. That is a different crate and outside this issue's scope; it wants its own issue.

Fixes #271

### DIFF
--- a/backends/hydrolysis/src/renderer.rs
+++ b/backends/hydrolysis/src/renderer.rs
@@ -202,6 +202,48 @@ pub(crate) use input::{
     text_editing,
 };
 
+/// The antialiasing vello runs for every Hydrolysis layer.
+///
+/// Multisampling rather than vello's analytic `Area` coverage, because `Area`
+/// cannot produce the same bytes twice. `path_count.wgsl` hands each segment
+/// its slot inside a tile with an `atomicAdd`, so the order the `fine` stage
+/// walks a tile's segments in is whatever order the dispatch happened to claim
+/// those slots; `fine` then accumulates coverage as `area += a * dy`, and
+/// floating-point addition is not associative. An unchanged scene therefore
+/// rasterizes to coverages an ULP apart between two renders, which quantizes to
+/// pixels 1/255 apart (#271). The multisampled fine stage accumulates winding
+/// numbers and sample masks with integer `atomicAdd`s instead, and integer
+/// addition does not care what order it is applied in, so the same scene always
+/// composites to the same bytes.
+pub(crate) const LAYER_ANTIALIASING: vello::AaConfig = vello::AaConfig::Msaa16;
+
+/// The fine-stage variants a Hydrolysis `vello::Renderer` compiles: exactly the
+/// one [`LAYER_ANTIALIASING`] asks for, so pipeline creation does not pay for
+/// variants no layer will ever select.
+pub(crate) const LAYER_ANTIALIASING_SUPPORT: vello::AaSupport =
+    antialiasing_support_for(LAYER_ANTIALIASING);
+
+/// The single-method [`vello::AaSupport`] that lets a renderer serve `config`.
+const fn antialiasing_support_for(config: vello::AaConfig) -> vello::AaSupport {
+    match config {
+        vello::AaConfig::Area => vello::AaSupport {
+            area: true,
+            msaa8: false,
+            msaa16: false,
+        },
+        vello::AaConfig::Msaa8 => vello::AaSupport {
+            area: false,
+            msaa8: true,
+            msaa16: false,
+        },
+        vello::AaConfig::Msaa16 => vello::AaSupport {
+            area: false,
+            msaa8: false,
+            msaa16: true,
+        },
+    }
+}
+
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) struct ContentSizeLimits {
     pub(crate) minimum: LayoutSize,
@@ -337,7 +379,7 @@ impl HydrolysisRenderer {
             device,
             vello::RendererOptions {
                 use_cpu: false,
-                antialiasing_support: vello::AaSupport::area_only(),
+                antialiasing_support: LAYER_ANTIALIASING_SUPPORT,
                 // Hydrolysis is the high-end, multi-core renderer: let vello parallelize
                 // pipeline initialization across all available cores instead of pinning
                 // it to a single thread.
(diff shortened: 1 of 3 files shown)


lexoliu:
Not merging yet: the Windows bench lane, which is green on `dev` every day this week, aborts with this change.

`Bench (windows-latest)` run 34089933202: `waterui_bench_stress_steady_redraw` dies after 1.5 s with `0xc0000005` (access violation) inside the stress example. Nothing else in the workspace changed on that lane, so the multisampled fine stage is the suspect on the runner's adapter (a software Dx12 adapter, most likely WARP). `Msaa16` needs about 4.6 KB of workgroup storage and takes a different `fine` variant; `Area` did not crash there.

Two options need a renderer-design decision: keep `Area` and land only the determinism test (with the vello per-tile segment ordering documented as the source), or keep `Msaa16` and first find why the Windows adapter faults. Until that decision this PR stays open.

lexoliu:
Closing with #271, which is closed as not planned: the one-pixel difference is accepted as a known issue, so the backend keeps vello's analytic `Area` anti-aliasing and this PR's switch to `Msaa16` is not wanted. The diagnosis it produced is recorded on #271. (It also aborted the Windows bench lane with an access violation on that runner's adapter, which would have needed solving first.)

Issue #239, opened 2026-09-02
graphics: shaderloom's precompiled SPIR-V flips Y, so every WaterUI shader renders upside down on Vulkan


lexoliu:
## Defect

Every shader WaterUI compiles through `shaderloom` renders **vertically flipped
on Vulkan**. Shaders wgpu compiles at runtime — Vello classic's and
`vello_hybrid`'s own — are unaffected, which is why this has stayed invisible:
the flip lands on WaterUI's blit, the Hydrolysis compositor, gradients, shader
surfaces and filters, and never on the vector content those composite.

Apple is unaffected (MetalLib artifacts take a different path), so it has never
shown up on a developer machine.

## Root cause

`shaderloom::build::compile_spirv` writes SPIR-V with naga's default writer
flags:

```rust
let words = spv::write_vec(module, info, &spv::Options::default(), None)
```

`spv::Options::default()` includes `WriterFlags::ADJUST_COORDINATE_SPACE`,
documented as *"Flip Y coordinate of `BuiltIn::Position` output"*. wgpu
deliberately does **not** use that flag for Vulkan: it maps WebGPU clip space
onto Vulkan's by submitting a negative-height viewport instead
(`wgpu-hal/src/vulkan/command.rs`). Passthrough SPIR-V is opaque to wgpu, so the
two flips compose and the frame comes out upside down.

`shaderloom::required_features` opportunistically enables `PASSTHROUGH_SHADERS`
wherever the adapter offers it, so Vulkan takes the precompiled path by default.

## Proof

A standalone program blits a source texture — first row red, last row blue —
through WaterUI's own `blit.wgsl`, with the *only* difference being how the
shader module is created. Same device, same pipeline, same draw:

| device | shader module | blitted row 0 |
| --- | --- | --- |
| Apple M2 Max, Metal | wgpu runtime WGSL | red — upright |
| Apple M2 Max, Metal | shaderloom precompiled | red — upright |
| llvmpipe 19.1.7, Vulkan | wgpu runtime WGSL | red — upright |
| llvmpipe 19.1.7, Vulkan | shaderloom precompiled | **blue — FLIPPED** |

Vello classic's `render_to_texture` was checked the same way and is upright on
both backends, so the vector content is fine and only the shaderloom-compiled
stage inverts it.

## Evidence in CI today

Every Linux snapshot this workspace uploads is upside down and has been for as
long as those artifacts have existed, because none of them asserted orientation:

[...]

PR #399, opened 2026-09-07
fix(graphics): drop the Mali negative-viewport wgpu fork

### LINKED ISSUE #240: Re-evaluate the Mali negative-viewport wgpu fork patch once shaderloom stops flipping Y (#239)
#239 shows shaderloom compiles SPIR-V with naga `WriterFlags::ADJUST_COORDINATE_SPACE`, which flips `BuiltIn::Position` Y on top of wgpu's own negative-height viewport flip, so every shaderloom-precompiled shader is upside down on Vulkan. The workspace still carries a `[patch.crates-io]` fork of wgpu (`lexoliu/wgpu` branch `mali-negative-viewport`) that flips Y in the vertex shader for the Pixel 9 Pro Mali r54p2 driver, on the finding that the driver "ignores the negative-height viewport". Those two facts overlap: a driver that ignores the negative viewport plus shaderloom's extra flip would produce exactly the observed inversion without any driver bug. Once shaderloom is fixed and the monorepo bumped, re-test the map / GpuSurface output on the Pixel 9 Pro (Tailscale `100.76.86.48:5555`) with the fork patch removed; drop the patch if it was compensating for #239.

lexoliu:
Fixes #240

Removes the `lexoliu/wgpu` `mali-negative-viewport` `[patch.crates-io]` block (wgpu, wgpu-core, wgpu-hal, wgpu-types, naga) so the nine wgpu crates come from crates.io again; `Cargo.lock` changes only their `source` lines.

**Evidence (Pixel 9 Pro, Android 16, Vulkan, Mali r54p2).** Same app (the HDR flame `GpuSurface` example, a fullscreen render-pass pipeline with bloom and tonemap), built and installed through `water run --platform android`, screenshot with `screencap`:

| run | orientation | logcat |
| --- | --- | --- |
| with fork (current dev) | upright | clean |
| without fork, first capture | upright; one transient frame showed noise on one triangle of the fullscreen pass, gone in the next two captures | clean, no `VK_ERROR`, no device loss |
| without fork, later captures (×2) | upright, identical to the fork run | clean |

The upside-down frames that motivated the fork were shaderloom's `ADJUST_COORDINATE_SPACE` flip stacked on wgpu's own negative-height viewport (#239); with #239 fixed the driver honours the viewport and the fork is dead weight.

Host: `cargo check --locked -p hydrolysis` passes on the new lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6970,7 +6970,8 @@ dependencies = [
 [[package]]
 name = "naga"
 version = "29.0.4"
-source = "git+https://github.com/lexoliu/wgpu?branch=mali-negative-viewport#3b9bdf1c8aa5193ca7766a31e665f1d689dfd2b6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2bf919621e7975acb27d881bae2fb993e0d45c8e0446e85e6272971e00dc8df"
 dependencies = [
  "arrayvec",
  "bit-set 0.9.1",
@@ -13690,7 +13691,8 @@ dependencies = [
 [[package]]
 name = "wgpu"
 version = "29.0.4"
-source = "git+https://github.com/lexoliu/wgpu?branch=mali-negative-viewport#3b9bdf1c8aa5193ca7766a31e665f1d689dfd2b6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76e8840e1ba2881d4cbb18d2147627a56af426ff064c0401eb0c8410c6325d07"
 dependencies = [
  "arrayvec",
  "bitflags 2.13.1",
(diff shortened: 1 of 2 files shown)


lexoliu:
Follow-up evidence on the same device, with the fork removed.

The map example is the repro #240 names, and it exercises the other GPU path: vello classic renders through compute and the result reaches the swapchain through WaterUI's shaderloom-compiled `blit.wgsl`, so it is the pipeline #239's flip actually landed on. Built and installed with `water run --platform android` from `examples/map` on this branch, then captured with `screencap`:

Manhattan renders north-up and correct — Upper West Side and Central Park at the top, Washington Square and Canal Street at the bottom, the Hudson on the left and the East River on the right. Every street label is upright and readable, none mirrored. `adb logcat -d | grep -iE 'vulkan|wgpu|device lost|VK_ERROR'` over that run has no `VK_ERROR` and no device loss; the only `device lost` matches are from Android's digital-car-key service and are unrelated.

Host verification on the new lock: `cargo check --locked -p hydrolysis`, `cargo clippy -p hydrolysis --all-targets -- -D warnings`, and `cargo nextest run -p waterui-graphics` (51 passed) are all clean.

PR #404, opened 2026-09-07
refactor(ffi): retire the CEF-specific input entry points for the GpuSurface carrier

### LINKED ISSUE #252: Retire the CEF-specific input entry points in favour of the GpuSurface input carrier

lexoliu:
Fixes #252

**FFI** (`ffi/src/components/platform/browser_cef.rs`): the CEF surface is created with `gpu_view_with_input(page)`, so the page's `CefInputGpuView` receives pointer, scroll, key, focus and IME composition through the generic carrier (`waterui_gpu_surface_send_input_event` and friends from #250). Deleted the eleven CEF-only exports that duplicated it (`set_focus`, `request_frame`, `set_viewport`, `pointer_move`, `pointer_button`, `scroll`, `key`, `commit_text`, `set_composition`, `finish_composition`, `cancel_composition`) plus `WuiCefPointerButton`, `WuiCefInputModifiers` and the dead range helper. Kept: `waterui_force_as_cef_webview`, `go_back`/`go_forward`, `prepare_macos_application`, `run_packaged_subprocess`, the state destructor `waterui_cef_surface_drop`, and `waterui_cef_surface_edit`.

**Why `edit` stays**: Edit-menu commands (copy/cut/paste/select all/undo/redo) reach a view as AppKit selectors, not key events, and `SurfaceInputEvent` has no edit-command variant. The right follow-up is an `Edit(SurfaceEditCommand)` variant on the carrier so the generic responder can implement those selectors for every GpuView; that would retire the last CEF-specific input export.

**Header**: regenerated with the nightly generator; both submodules carry the copy.

**Apple** (`backends/apple` d2973512): `CefSurfaceView` drops its own `NSTextInputClient`, tracking area, key/scroll/pointer interception, display link and viewport bookkeeping (583 → 96 lines) and hosts a `WuiGpuSurface`, whose responder handles input; back/forward, the Edit-menu selectors and `deinit` remain.

**Android** (`backends/android` 614ed10e): header only.

Verified: `swift build` of the Apple package (WaterUICEF, WaterUICefWebView, WaterUIChromium compile clean), `cargo check -p waterui-ffi`, `cargo clippy -p waterui-browser-cef --all-targets -- -D warnings`, the CLI scaffold pin test. The `--all-features` lint of `waterui-ffi` fails on `dev` already because `c-api` and `android-jni` are mutually exclusive by design; unchanged here. The macOS real-engine check (typing, scrolling, IME through the generic path) is the acceptance criterion from the issue and is required before merge.

---

**Acceptance run (macOS, `examples/chromium`, built from this branch).** First launch panicked on the first CEF paint: `RefCell already mutably borrowed` at `ffi/src/components/visual/gpu_surface.rs:547` (`waterui_gpu_surface_measure`). Backtrace: `CefGpuView::render` → `page.pump()` → `OnAcceleratedPaint` → `MacFrameSink::import` → mailbox waker → `RedrawHandle::request_redraw` → Swift `handleExternalRedrawRequest` → `handleRedrawRequest` → `completeSetupIfReady`/`takeMeasurementInvalidation` → `waterui_gpu_surface_measure`, all while Rust's `with_semantic_mut` render borrow was live. The old `CefSurfaceView` never took this path because it did not host the generic `WuiGpuSurface` redraw plumbing. Fixed in `backends/apple` 5bb1cc89: the render state tracks the in-progress render call, parks a wake that arrives during it and replays it after the call returns (`withRenderInProgress`). Relaunched: first paint received, no panic, the page (waterui.dev) renders in the surface with the CDP controls beside it (window capture read by eye).

Still to do before merge: the interactive half of the acceptance (typing, scrolling, IME composition through the generic responder). That requires application-control tooling, which was unavailable during this verification.

Observed and filed separately: #406 (CEF pump trips the main-thread frame-budget guard every few seconds while idle; predates this change).

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -25,9 +25,9 @@ waterui-preview-version = "0.1.0"
 waterui-preview-protocol-version = "0.1.0"
 android-kotlin-version = "2.3.21"
 apple-backend-url = "https://github.com/water-rs/apple-backend.git"
-apple-backend-commit = "180dfaa559eea9ef6434c1c23d7d7c4b8c07cc7e"
+apple-backend-commit = "5bb1cc890acf802c6e4e2e6130b1c91531743af7"
 android-backend-url = "https://github.com/water-rs/android-backend.git"
-android-backend-commit = "e4786d52a3ed7ce969311ac437725ad93040adfa"
+android-backend-commit = "438c232d5cbf61f35795b6faf715f45c80094da7"
 
 [[bin]]
 name = "water"
--- a/ffi/src/components/platform/browser_cef.rs
+++ b/ffi/src/components/platform/browser_cef.rs
@@ -5,8 +5,7 @@ use std::{any::Any, fmt};
 #[cfg(any(feature = "webview-cef", feature = "cef-header"))]
 use waterui_browser_cef::CefWebViewHandle;
 use waterui_browser_cef::{
-    CefInputModifiers, CefKeyInput, CefPageHandle, CefPointerButton, CefRuntime,
-    CefRuntimeConfiguration, CefTextRange, gpu_view,
+    CefPageHandle, CefRuntime, CefRuntimeConfiguration, gpu_view_with_input,
 };
 #[cfg(any(feature = "chromium", feature = "cef-header"))]
 use waterui_chromium::{ChromiumView, PageMode};
@@ -18,7 +17,7 @@ use waterui_webview::WebView;
 #[cfg(any(feature = "webview-cef", feature = "cef-header"))]
 use crate::WuiAnyView;
 use crate::components::visual::gpu_surface::WuiGpuSurface;
-use crate::{IntoFFI, IntoRust, WuiStr};
+use crate::{IntoFFI, IntoRust};
 
 /// Installs one process-owned CEF runtime and the selected public controllers.
 pub(crate) fn configure_environment(env: &mut Environment) {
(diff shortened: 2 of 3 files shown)


lexoliu:
Live acceptance on macOS, `examples/chromium` built from this branch (`020489da5`, apple `5bb1cc89`).

**The new carrier path is provably active.** Attaching a debugger to the running app and dumping the window's view hierarchy shows the browser surface as

```
WaterUI.WuiAccessibilityRole
  WaterUI.WuiGpuSurface            f=(0,0,450,600)
    WaterUI.WuiGpuSurfaceInputResponder  f=(0,0,450,600)
```

That responder is installed only when `waterui_gpu_surface_wants_input_events` answers true, which is only true because the surface is now built with `gpu_view_with_input(page)` — `CefInputGpuView::wants_input_events` returns `true`, while the bare `CefGpuView` this branch replaced took the default `false`. On `dev` no responder exists there. So the generic input carrier is wired to the CEF page, sized to the whole surface, in a live app.

**Rendering is unaffected**: waterui.dev paints in the surface, no panic (the first launch did panic on the first accelerated paint; that was a re-entrancy bug fixed in apple `5bb1cc89`, described above).

**What this example does not cover.** `ChromiumView::body` prefers a `Hook<ChromiumView>` when the application links the CEF runtime, and that hook yields the GPU surface above — so `ChromiumComponent`/`CefSurfaceView`, the Swift half rewritten here, is the *other* configuration: an Apple backend that hosts Chromium natively by type (`waterui_chromium_id`). It compiles (`swift build` clean, all three CEF targets) and its remaining duties are back/forward, the Edit-menu selectors and the state destructor, but it is not instantiated by this example. Exercising it, and typing/scrolling/IME through the responder as a person would, needs the interactive channel; that half of the issue's acceptance criterion is still open.

PR #415, opened 2026-09-07
feat(codes): draw QR codes as a self-drawn scene component

### LINKED ISSUE #113: QR code

lexoliu:
No platform ships a QR primitive — a code is a picture everywhere — so `AGENTS.md` Principle 4 settles the architecture, and `waterui-qr` is a Rust-side self-drawn component with **zero new FFI types**. `qrcode` supplies the module grid and the grid is drawn through `Scene2D`, so a code renders on whichever engine the backend supplies (classic compute, the CPU/GPU split engine, dew's CPU scene) and stays vector sharp instead of being rasterized once into a bitmap.

## Encoder

`qrcode` 0.14.1 over `fast_qr` 0.14, `default-features = false`:

- **Licence.** MIT OR Apache-2.0, matching the workspace's own dual licence. `fast_qr` is MIT only, which narrows what WaterUI can offer downstream.
- **Exactly the API needed.** `QrCode::with_error_correction_level(…)` then `into_colors()` is the module matrix and nothing else; with default features off, the image, SVG and PIC renderers (and the `image` crate behind the first) are unreachable, leaving zero transitive dependencies. `fast_qr` is equally clean on that axis but stores every symbol as an inline 177×177 array, so the smallest code carries the largest symbol's storage.
- **Typed failure.** `QrError` is a real enum, so a payload past the largest symbol is a typed error rather than a panic or a silently smaller code.

## Snapping

One module is a whole number of units on a side and the block sits at a whole-unit origin, so every module boundary is exact. A boundary that falls inside a pixel is antialiased into a grey seam, and a decoder's binarizer then has to guess which side that seam belongs to — at small sizes it guesses wrong on most rows, which is more damage than the error correction was sized for. The space snapped to is the one `build_scene` is handed: the surface's own pixel grid under `GpuSurface`, the window's logical grid (the device pixel grid at every integer scale factor) where a backend merges the scene into its own. Whatever the box does not divide evenly is left as extra quiet zone, the one thing around a symbol that may grow freely.

The quiet zone itself — four modules, per ISO/IEC 18004 — is part of the drawing rather than something a caller must remember.

## Colours, and why the pair is ordered

Style is an attribute (Principle 1): correction level, quiet zone, module size and both colours are methods on the view, and the colours take `impl IntoComputed<Color>`.

The defaults are the theme's `Foreground` and `Surface` (Principle 6), but the **darker of the two always draws the modules**. A decoder looks for dark modules on a light ground, and the detectors in wide use — `quirc`, ZXing, and `rqrr`, which this PR's tests decode with — do not search for the reflectance-reversed form at all, so taking the pair verbatim would hand a dark-mode application a code that looks right and does not scan. Ordering by perceptual lightness keeps a themed application's own near-black and near-white rather than a hardcoded `#000` on `#fff`, and keeps the polarity a decoder needs across a colour-scheme switch. Naming either colour hands the judgement back to the caller.

## Tests

The round trip is the load-bearing one: the component is rendered offscreen through `waterui-testing`'s existing GPU-backed session and the pixels are decoded back with `rqrr`. Every way a drawn code stops being readable leaves a picture that still looks like a QR code, so nothing short of a decoder is evidence.

- two pixels a module and eleven pixels a module, both decode;
- a payload change through a `Binding` — the *new* payload comes back out of the frame;
- the dark theme decodes, and a deliberately inverted code is **not found**, which is the evidence behind the colour ordering rather than a claim about it;
- `tests/e2e_semantics.rs` asserts the accessibility node carries the payload, that an application label wins over it, that it follows the signal, and that the code stays square in a stack;
- an over-long payload is a `QrError::PayloadTooLong`, and finder-pattern geometry pins that the grid is indexed the way `is_dark` claims.

Frames land in WaterUI's canonical artifact layout, so the run that proves a code decodes also leaves the image to look at. Both were read directly: quiet zone present, finder patterns square with clean light rings, hard module edges with no bleeding even at two pixels a module.

## Also in this diff

`ResolvedColor::to_peniko` moves into `waterui-graphics`, where the conversion every scene-drawing component needs belongs. `waterui-math` carried a private copy — pre-existing duplication, not a defect — and now uses the shared one instead of this PR adding a third.

Fixes #113

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4004,6 +4004,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "g2gen"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5a7e0eb46f83a20260b850117d204366674e85d3a908d90865c78df9a6b1dfc"
+dependencies = [
+ "g2poly",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
+]
+
+[[package]]
+name = "g2p"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "539e2644c030d3bf4cd208cb842d2ce2f80e82e6e8472390bcef83ceba0d80ad"
+dependencies = [
+ "g2gen",
+ "g2poly",
+]
+
+[[package]]
+name = "g2poly"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "312d2295c7302019c395cfb90dacd00a82a2eabd700429bba9c7a3f38dbbe11b"
+
+[[package]]
 name = "gallery-example"
 version = "0.1.0"
 dependencies = [
@@ -4886,7 +4914,7 @@ dependencies = [
  "jiff",
  "js-sys",
  "keyboard-types",
- "lru",
+ "lru 0.16.4",
  "nami",
  "native-executor",
  "objc2 0.6.4",
@@ -6441,6 +6469,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f66e8d5d03f609abc3a39e6f08e4164ebf1447a732906d39eb9b99b7919ef39"
 dependencies = [
  "hashbrown 0.16.1",
+]
+
+[[package]]
+name = "lru"
+version = "0.18.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ff9840bcc50b71349309900da0ce7279aa336ae71d73250b07998932c7d97c25"
+dependencies = [
+ "hashbrown 0.17.1",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


lexoliu:
Closing: this duplicates a component WaterUI already ships.

`waterui-barcode` 0.1.0 is a workspace dependency (`Cargo.toml:178`), published from this project and developed in `water-rs/barcode`. It exposes `Barcode::qr(content)` with `.size()`, alongside `code128` and the rest of `BarcodeSymbology`, and it carries the accessibility label per symbology (`src/view.rs:69-71`). `examples/barcode` uses it and compiles against it today; the user-facing skill documents it at `references/media.md:175-177`. So the semantic component exists, is reachable from application code, and is documented.

It is also already a self-drawn Scene2D component: `water-rs/barcode` moved off the wgpu pipeline in `73dc270 feat!: draw barcodes through Scene2D instead of a wgpu pipeline` and gained intrinsic sizing in `015883c`. The published 0.1.0 predates both, so the way to get Scene2D barcodes into this workspace is a release of that crate and a version bump here — not a second crate.

Beyond the duplication, `waterui-qr` as a separate type is the parallel-type shape Principle 1 rules out: which symbology a code uses is an attribute of one semantic component, which is exactly what `BarcodeSymbology` already is. Adding a QR-only type next to `Barcode` is the `CheckboxToggle` mistake in another costume.

The work here that has no counterpart upstream is the evidence: rendering the component offscreen and decoding the pixels back with `rqrr`, the module snapping argument, and the ordering of the colour pair by lightness so a dark-mode application still produces a scannable polarity. That is the right test for a barcode and it belongs in `water-rs/barcode`.

No criticism of the work — the crate is invisible from inside this repository unless you go looking at `Cargo.toml`, which is a gap in this repo's guidance rather than in the PR.

Issue #297, opened 2026-09-03
mermaid: propose the typed layout projection to upstream merman and drop the fork pin


lexoliu:
`waterui-mermaid` depends on `merman-core` / `merman-render` from our fork
`water-rs/merman`, branch `typed-layout-projection`, pinned in the workspace `Cargo.toml` at
rev `a6f41d47`. The fork carries one commit on top of upstream `Latias94/merman` `main`, in
`crates/merman-render/src/family.rs` only:

- `LayoutProjection` becomes `pub`
- `BuiltinFamilyArtifact::layout_projection` becomes `pub`
- `FamilyRenderArtifact::layout()` is added, returning `LayoutProjection<'_>`

Two visibility keywords and a 12-line accessor. No new types, no serde changes, no behaviour
change; `layout_json()` is untouched.

## Why the fork exists

Upstream's only public exit from the layout stage is `layout_json()`, and it is lossy by
design: fields marked `#[serde(skip)]` never reach a JSON consumer. For sequence diagrams
that drops `SequenceDiagramLayout::block_layouts_by_id`, the extents of every `loop` / `alt` /
`opt` fragment box, and `SequenceBlockLayout` does not derive `Serialize` at all. The probe
in #217 shows the same diagram with zero fragment boxes through JSON and two through
`layout()`. WaterUI draws the typed geometry itself (labels are real `text()` views, colours
are theme tokens), so it needs the projection, not the SVG.

## The upstream ask

Propose exactly that commit to `Latias94/merman` as a PR titled "Expose the typed layout
projection alongside the compatibility JSON". The body is drafted and has been reviewed
against the code; it offers to add a doc example on `layout()` and a test asserting the
projection is reachable if the maintainer wants either.

Held deliberately, not filed: anything sent upstream goes out under the maintainer's own
name and is hand-verified by them first. The decision on 2026-09-03 was to keep using the
fork for now and revisit once the design has been exercised further.

## When to revisit

- Upstream `family.rs` is refactored and the fork has to be rebased more than trivially.
- A second consumer field turns out to be behind `#[serde(skip)]` (nine are; #217 lists them).
- A merman release we want to move to makes carrying the patch a recurring cost.

## What filing involves

1. Rebase `typed-layout-projection` onto current upstream `main`; `cargo check -p merman-render`.
2. Optionally add the doc example and reachability test to the fork branch first.
3. The maintainer reviews the drafted text, then files the PR from their account.
[...]

Issue #217, opened 2026-09-02
waterui-mermaid reads layout geometry as JSON, losing sequence fragment boxes


lexoliu:
`waterui-mermaid` (#215) reads its layout geometry back out of `merman-render` as JSON, and
should not have to.

`merman-render` computes the geometry into typed values — `LayoutNode`, `LayoutEdge`,
`LayoutCluster`, `Bounds`, `FlowchartLayout`, `SequenceDiagramLayout` and the rest, all
`pub` in its `src/model.rs`. What is not public is the way to reach them:

```rust
// merman-render 0.8.0-alpha.5, src/family.rs
:328    enum LayoutProjection<'a> { .. }                        // not pub
:506    fn layout_projection(&self) -> LayoutProjection<'_>     // not pub
:775    pub fn layout_json(&self) -> Result<serde_json::Value>  // the only public exit
```

and `layout_json`'s body is `serde_json::to_value(self.family.layout_projection())`. The JSON
is not an integration contract anyone designed for consumers; it is the parity-debugging
projection used to diff merman's geometry against upstream Mermaid fixtures — the
`#[serde(rename = "FlowchartV2")]` style attributes on the projection are that comparison's
naming, not ours.

Two costs, one of them real:

1. A serialize/deserialize round trip per layout. At diagram scale this is not a bottleneck.
2. **Nine fields carry `#[serde(skip)]` and do not survive the trip.** One of them lands
   inside this issue's scope:

   ```rust
   pub struct SequenceDiagramLayout {
       ..
       /// This remains outside compatibility layout JSON because it is renderer-owned phase data.
       #[serde(skip)]
       pub block_layouts_by_id: FxHashMap<String, SequenceBlockLayout>,
   }
   ```

   `SequenceBlockLayout { start_y, stop_y, section_ys_by_id }` is the vertical extent of a
   sequence diagram's `loop` / `alt` / `opt` fragment boxes. It is not `Serialize` at all, so
   the fragments have to be re-derived from the semantic model plus each message's laid-out
   `y` — recomputing something merman already computed exactly.

Flowchart is unaffected in practice: the only skipped fields it loses are `label_width` /
`label_height`, which our own `TextMeasurer` produced in the first place.

Until this is resolved, keep the JSON strictly behind our own `DiagramLayout` boundary type,
so the renderer never names a `serde_json::Value` and switching the source is an internal
change.

**The upstream ask, for the record and not to be filed without approval:** make
[...]

PR #417, opened 2026-09-07
fix(mermaid): measure diagrams with the shared font collection

### LINKED ISSUE #409: mermaid: a fourth FontContext behind merman-render's Arc<dyn TextMeasurer + Send + Sync>
Found while landing #407 (#230). `components/data/mermaid/src/measure.rs` constructs its own `parley::FontContext` because `merman-render` requires `Arc<dyn TextMeasurer + Send + Sync>`, which the single-threaded `waterui_text::collection::FontCollection` (`Rc<RefCell<FontContext>>`) cannot satisfy. Either merman's measurer contract loosens to a main-thread trait object (it is only ever called from layout on the UI thread), or the mermaid component shapes through a measurer that borrows the shared collection during layout and hands merman a precomputed metrics table. Design decision first; then the fourth per-component font enumeration goes away.

lexoliu:
Fixes #409

`components/data/mermaid/src/measure.rs` built a `parley::FontContext` of its own — the fourth in the workspace after #407 moved three components onto `waterui_text::FontCollection` — and wrapped it in a `Mutex` for the sole purpose of satisfying `merman-render`'s `Arc<dyn TextMeasurer + Send + Sync>`. Neither the second font enumeration nor the lock had a reason to exist: a diagram is laid out on the thread that owns the view, from the faces the rest of the window already shapes with.

## What changed

The measurer borrows the collection the host installed, read out of the `Environment` the way `waterui-math` reads it, and satisfies merman's bound with `waterui_core::MainThreadBound` instead of a lock. The wrapper is how the type system states "this measurer never leaves the thread it was built on", and it asserts exactly that on every access — a fail-fast net rather than an uncontended mutex taken on every `measure` call. A host that installed no collection gets the collection's own message naming its responsibility; the component builds no fallback, because that is the duplication being removed.

`Mermaid::body` now reads `FontCollection::from_env(env)` and hands it to `engine::render`. The crate no longer enables `parley/system`; it takes `waterui-text`'s `system-fonts` feature instead, exactly as `waterui-math` does, so a build carrying diagrams can still reach real faces on a host that owns no font stack.

## Why merman's contract was left alone

Loosening `Arc<dyn TextMeasurer + Send + Sync>` in our fork was the other candidate, and it does not hold. The bound is not decoration inside `merman-render` — nothing there moves the measurer across a thread, the only `std::thread` uses in that crate's `src` are in `#[cfg(test)]` modules — but it is load-bearing one crate up. `merman-cli` renders the diagrams of one Markdown file across a `rayon` pool, sharing a single `&MarkdownRenderContext` (`crates/merman-cli/src/render/markdown_export.rs:57-67`) whose `&PreparedGraphicalRender` holds the `RenderEnvironment` the `TextMeasurementPolicy` lives in. Dropping `Send + Sync` would take `parallel-markdown` away from our own fork.

It would also cost us upstream. `water-rs/merman` diverges from upstream by exactly one commit today — the purely additive `feat(render): expose typed layout geometry alongside the compatibility JSON` that #217 needed. A second, breaking commit that removes a capability from a headless renderer used by CLI, node and wasm hosts is a much harder thing to propose (#297).

`MainThreadBound` costs one thread-id comparison per measurement and no divergence at all.

## A pre-existing defect the new test found

The measurer honoured Mermaid's CSS `font-family` — `"trebuchet ms", verdana, arial, sans-serif` by default — while the label layer paints labels in the application's own font. So a box was measured in one family and painted in another, which is the very thing this module exists to prevent: the edge label `yes` shapes to 24.90pt inside a box reserved at 23.09pt. This was pre-existing, not introduced here; the new layout test is what made it visible.

Diagram typography comes from the application, the same way diagram colours come from theme tokens and never from Mermaid's CSS themes, so the family is no longer pushed. Weight and style still are: they only ever make a reserved box larger.

## Tests

- `every_reserved_box_holds_the_text_it_will_be_painted_with` — for a flowchart and a sequence diagram, every label's reserved box is checked against what the same collection shapes that label to, at the size and weight the label layer paints it with. This is the property the module docs claim, and it is the test that caught the family mismatch above.
- `two_diagrams_in_one_app_share_one_collection` — pointer identity, mirroring the `waterui-text` test #407 added.
- `the_installed_collection_is_what_measures` — a diagram measured against a collection with no faces does not size like one measured against the system's, so the test above cannot pass against a measurer that ignores its collection.
- `a_diagram_without_a_collection_names_the_host` — drawing with no collection installed panics with the collection's own message.

## Verification

`cargo check -p waterui-mermaid --all-targets`, `cargo clippy -p waterui-mermaid --all-targets -- -D warnings` (clean), `cargo nextest run -p waterui-mermaid` (21/21), `cargo test --doc -p waterui-mermaid`, `cargo check -p waterui-text -p waterui-math`, `rustfmt --edition 2024` on the changed files.

The exported review PNGs were compared before and after: the diagrams are identical apart from sub-pixel box metrics.

### DIFF
--- a/components/data/mermaid/Cargo.toml
+++ b/components/data/mermaid/Cargo.toml
@@ -15,7 +15,11 @@ waterui-core.workspace = true
 waterui-canvas.workspace = true
 waterui-graphics.workspace = true
 waterui-layout.workspace = true
-waterui-text.workspace = true
+# `system-fonts` is what lets a host with no font stack of its own — the native
+# Apple, Android and GTK backends — install a collection at all. A diagram is
+# measured from real faces, so a build carrying this component must be able to
+# reach them.
+waterui-text = { workspace = true, features = ["system-fonts"] }
 waterui-str.workspace = true
 nami.workspace = true
 merman-core.workspace = true
@@ -23,10 +27,11 @@ merman-render.workspace = true
 # Only to read Mermaid's own `fontSize` out of the effective config, which
 # `merman-core` exposes as a JSON value.
 serde_json.workspace = true
-# The host text measurement `merman-render` lays diagrams out with. Same engine
-# and the same `system` font source `waterui-canvas` draws labels with, so the
-# boxes layout reserves are the boxes the glyphs land in.
-parley = { workspace = true, features = ["system"] }
+# The host text measurement `merman-render` lays diagrams out with. The faces
+# come from the application's shared `FontCollection`, which this crate only
+# reads: `parley` is here for the shaping the collection is handed to, not to
+# discover fonts.
+parley = { workspace = true, features = ["std"] }
 thiserror.workspace = true
 
 [dev-dependencies]
--- a/components/data/mermaid/src/engine.rs
+++ b/components/data/mermaid/src/engine.rs
@@ -697,13 +697,10 @@ fn lifeline_span(lifelines: &[Lifeline], actors: &[String]) -> (f32, f32) {
 
 #[cfg(test)]
 mod tests {
-    use merman_render::text::TextStyle;
-    use parley::FontWeight;
     use waterui_core::Environment;
     use waterui_text::FontCollection;
 
     use super::render;
-    use crate::layout::{Emphasis, Label};
     use crate::measure;
 
     const FLOWCHART: &str = "\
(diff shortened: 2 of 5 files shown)


lexoliu:
Rebuilt on current `dev` and added the fix for the CI failure.

The Linux runner was right and the test was catching a real defect: the box `merman` reserved for the subgraph title `Ingest` was 49.469pt wide while the same collection shaped that text to 49.718pt. Measurement was not the problem — painting was. `merman` reserves a title's box from a measurement at the diagram's font size, in the weight its own style asks for, and the label layer then drew titles at seven eighths of that size in **bold**. Two metrics for one piece of text; whether the smaller box could hold the bolder glyphs was a property of the face, so macOS fit and the CI image did not.

`measure::label_style` is now the single style a label is measured and painted with — a function of the diagram's font size alone — and both the paint layer and the test read it instead of each restating what a title should look like, so the two cannot drift again. Prominence became a colour, which cannot change an advance: a fragment's guard condition draws in `MutedForeground` rather than a step smaller. Subgraph titles and participant names lose the bold, which is also what Mermaid itself draws.

Verified by reading the exported diagrams: subgraph titles sit inside their frames at body size, the guards beside `loop` and `alt` are legibly quieter, every label is inside its box. `clippy --all-targets -D warnings` clean, 25/25 tests, doctests pass.

PR #437, opened 2026-09-08
fix(hydrolysis): shape non-Latin scripts through the platform's fallback

### LINKED ISSUE #426: hydrolysis: CJK text renders as tofu although the system has CJK faces
Found while landing the Avatar component (#93). The gallery's avatar for `山田 太郎` shows two empty boxes where its monogram `山太` should be. The string itself is correct — the monogram derivation is unit-tested on that exact name, and `"Ольга Ладыженская"` (Cyrillic) renders fine — so the failure is in shaping or font selection, not in the text.

This is on macOS, where the system unquestionably has CJK faces installed, so it is not "no font is available". Two candidate causes worth separating before fixing:

- the collection resolves one family and does not fall back per script when the primary face has no glyph for a cluster, so anything outside that face's coverage becomes `.notdef`;
- the collection is built from a fixed family list that never asks the platform for a face covering the run's script.

Either way the observable is the same and it affects every non-Latin script, not only CJK — an application whose content is Chinese, Japanese, Korean, Arabic, Hebrew, Thai or Devanagari gets boxes from the self-drawn renderer while the native backends draw it correctly, which makes hydrolysis unusable for most of the world's text.

The avatar with the Japanese name is deliberately left in the gallery as the reproduction rather than swapped for a Latin one. Note that a fix which bundles a face has a package-size dimension (Principle 5) and would re-baseline the existing snapshots, so measure before choosing; a fix that adds script fallback over the platform's own faces has neither cost and is the first thing to check.

lexoliu:
Fixes #426

## Which of the two candidate causes it was

The second, and it is in the **headless test host**, not in the shipping runner.

`deterministic_test_fonts` built its `fontique::Collection` with `system_fonts: false` and registered four bundled Roboto files. With discovery off there is no platform backend to ask, so `Collection::fallback_families` can only return what was installed by hand — and nothing installs a fallback for a script the bundled faces do not carry. Every cluster outside Roboto's coverage therefore shaped to `.notdef`, which is glyph 0, which draws as the tofu box. Roboto carries Cyrillic, which is exactly why `Ольга Ладыженская` → `ОЛ` looked fine in the same gallery that drew `山田 太郎` → `山太` as two empty squares.

`native_resource_fonts`, the collection the shipping runner installs, starts from `parley::FontContext::new()` with discovery on and has always reached the platform's own fallback (CoreText on Apple, fontconfig on Linux, DirectWrite on Windows). Measured against parley 0.11.1 on this machine, all six scripts resolve to a real face with zero `.notdef` there.

## The fix

Discovery goes back on in the test host, and the determinism it was switched off for is provided by the mechanism that actually provides it: `ResourceFontFamilies::install` pins `SansSerif` / `UiSansSerif` / `SystemUi` to the bundled Roboto, and a family the collection registered itself is matched ahead of any system family of the same generic. Latin and Cyrillic still shape through Roboto on every runner, byte for byte; only what Roboto cannot map reaches the platform's fallback — the same path the runner takes.

No font is bundled and no snapshot is re-baselined: the full test suites of `hydrolysis`, `waterui`, `waterui-testing`, `waterui-text`, `hydrolysis-m3` and every component crate with offscreen tests pass unchanged.

## Evidence

Three tests in `runner::fonts` state it without looking at a pixel, shaping through the renderer's own `TextMeasureService`:

- no sample of Latin, Cyrillic, Han, Hangul, Arabic, Hebrew, Thai or Devanagari shapes to glyph 0;
- the two scripts Roboto carries still resolve to a bundled face (the determinism half);
- the six it does not resolve to a face that is *not* bundled, so the test above cannot pass on a collection that quietly stopped registering anything.

All three fail on `dev`: reverting just `system_fonts` reports `4 of 5 glyphs in the Han sample 山田 太郎 are .notdef`.

`tests/text_scripts.rs` renders the same samples through the existing offscreen snapshot path for direct image review, light and dark. Read by eye on macOS: every row draws its own script, Arabic joins and orders right-to-left, Hebrew orders right-to-left, Thai stacks its marks and Devanagari forms its conjunct. The avatar gallery — the reported reproduction, left untouched in the tree — now draws `山太`.

## CI

`fonts-noto-core` and `fonts-noto-cjk` are added to the Linux runner's package list. The tests fall back to the platform's faces by design, so a runner with none installed would fail them for want of a font rather than for want of a fix.

### DIFF
--- a/.github/actions/setup-linux-deps/action.yml
+++ b/.github/actions/setup-linux-deps/action.yml
@@ -44,7 +44,13 @@ runs:
         # (xcap -> waterkit-screen pulls it in under --all-features). It has so
         # far arrived only as a transitive dependency of libgtk-4-dev; declare
         # what the workspace actually links rather than inheriting it.
-        packages="nasm libasound2-dev libva-dev libfontconfig1-dev libgbm-dev libegl-dev libxcb1-dev libglib2.0-dev libpango1.0-dev libgdk-pixbuf-2.0-dev libgtk-4-dev libpipewire-0.3-dev libudev-dev"
+        # fonts-noto-cjk and fonts-noto-core are faces, not headers: the
+        # self-drawn renderer shapes a script the bundled Roboto does not cover
+        # through the platform's own fallback, exactly as it does on a user's
+        # machine, so a runner with no CJK, Arabic, Hebrew, Thai or Devanagari
+        # face installed makes that text shape to `.notdef` and the coverage
+        # tests fail for want of a font rather than for want of a fix.
+        packages="nasm libasound2-dev libva-dev libfontconfig1-dev libgbm-dev libegl-dev libxcb1-dev libglib2.0-dev libpango1.0-dev libgdk-pixbuf-2.0-dev libgtk-4-dev libpipewire-0.3-dev libudev-dev fonts-noto-core fonts-noto-cjk"
         if [ "${{ inputs.gpu }}" = "true" ]; then
           packages="$packages mesa-vulkan-drivers libvulkan1"
         fi
--- a/backends/dew/src/views/navigation.rs
+++ b/backends/dew/src/views/navigation.rs
@@ -732,21 +732,33 @@ fn render_destination(
     });
     let bottom = entry.chrome.bottom_layout(renderer.state_cell());
 
-    let bar_height = bar.as_ref().map_or(0.0, |layout| layout.height);
-    let bottom_height = bottom.as_ref().map_or(0.0, |(height, _)| *height);
+    // Where the chrome stops and the content starts is one edge, so it is
+    // computed once — on layout's `f32` grid, the coarser of the two the
+    // backend works in. Painting the bar to its `f64` height and starting the
+    // content at that height narrowed to `f32` describes two edges up to half
+    // an ULP apart, which is a hairline of window background under the bar or a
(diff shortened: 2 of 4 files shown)


lexoliu:
Pushed `1fde83db0`: this branch was red on `Test / ubuntu-latest` and the failure is real, not flake.

`a_stack_draws_its_root_under_a_bar` and `a_bar_places_every_part_it_declares` assert that a navigation bar's bottom edge *is* the content's top edge. Dew computed that edge twice — the chrome painted to `bounds.y0 + bar_height` in kurbo's `f64`, the content starting at the same sum narrowed to layout's `f32` — so the two agree only when the sum happens to be exactly representable in `f32`. Every bar height reached so far was; the fallback in this branch moves a title's measured height off that grid, and the assertion caught the split at once:

```
  left: Rect { x0: 0.0, y0: 65.68800354003906, x1: 240.0, y1: 240.0 }
 right: Rect { x0: 0.0, y0: 65.6879997253418,  x1: 240.0, y1: 240.0 }
```

Half an `f32` ULP of window background showing under the bar, or half an ULP of content painted beneath it. Both seams are now decided once, on the coarser grid, and the chrome rects and the content frame read those values. Pre-existing defect, filed as #441 and fixed here because this branch is what makes it reachable.

Issue #432, opened 2026-09-08
Drop the waterui-canvas git pin once a release carrying the arc_to fix ships


lexoliu:
The fix for #418 pins `waterui-canvas` to a git revision:

```toml
waterui-canvas = { git = "https://github.com/water-rs/canvas", rev = "ed4cecc706f69f381555f3fe193134da6e0ff7fe" }
```

That revision is water-rs/canvas#5, merged, which restores the `Path::arc_to` implementation this repository fixed in #228 and adds `Path::bounding_box`. The published 0.1.0 predates both.

A git dependency is not a resting place: it cannot be published from, it pins a revision rather than a version range, and it hides from the dependency tooling that watches for updates. Once a `waterui-canvas` release carrying that commit is on crates.io, replace the pin with the version and delete the comment above it.

Releasing that crate is a publish decision and belongs to @lexoliu; this issue exists so the pin is not forgotten once it happens.

lexoliu:
The release side is ready to go: water-rs/canvas#6 promotes `dev` to `main`, which is what makes release-plz cut the version carrying `Path::arc_to`. Published 0.1.0 predates the fix; `dev` has carried it since water-rs/canvas#5 (`ed4cecc`, which is exactly what this workspace pins).

Merging that PR publishes and requires release approval. When it lands, this issue closes by turning both `waterui-canvas` entries in the root `Cargo.toml` — `[workspace.dependencies]` and the `[patch]` section — back into a version requirement and deleting the two comment blocks that explain the pin.

Issue #431, opened 2026-09-08
Fixes made in-tree before a crate was split out may be missing from its standalone repository


lexoliu:
#418 turned out to be a lost fix rather than a new bug, and the way it was lost is repeatable.

`Path::arc_to` was fixed in this repository in `ef11533` (#228) on 2 September. `water-rs/canvas` had already been forked from the in-tree copy before that date, so it never carried the fix. On 4 September `cba85e1c` deleted the in-tree crate and took `waterui-canvas` 0.1.0 from crates.io, and the fix silently left the build with it. Nothing failed: the crate compiled, the API was identical, and the only symptom was that flowchart edges with a non-right-angle turn drew off into space — which took a rendered picture to notice, three days later.

Every crate that left this tree is exposed to the same thing, and the window is the gap between the fork point and the deletion commit: `waterui-barcode`, `waterui-chart`, `waterui-map-gpu`, `waterui-canvas`, `waterui-particle`, `waterui-image`, `waterui-video-gpu`, `waterui-visualizer`, `filtrate`, `shaderloom`, `nami`.

Work: for each, find the commit that removed the in-tree copy, diff the tree at its parent against the standalone repository at the revision the deletion pinned, and report every difference that is not a packaging change. A difference is either a fix that never reached the standalone repository — which needs porting there — or a deliberate divergence, which should be written down. The audit is mechanical and worth doing once now, and the same diff is worth taking before any future split.

The wider point for splits from here on: deleting the in-tree copy and pinning the published crate is not complete until that diff is empty.

lexoliu:
Audit done. Method: for each crate, take the in-tree source at the parent of the commit that deleted it, and diff it against the crate the workspace actually consumes — the published version vendored in the cargo registry — restricted to `src/`. That comparison is the one that matters, because the published artifact is what ships.

| crate | in-tree path, deleted | verdict |
|---|---|---|
| `waterui-canvas` 0.1.0 | `components/visual/canvas`, `cba85e1c1` | **functional loss** — `Path::arc_to` |
| `waterui-image` 0.3.0 | `components/visual/image`, `e1c64b1f2` | latent build loss — `#![recursion_limit = "256"]` |
| `waterui-visualizer` 0.3.0 | `components/multimedia/visualizer`, `05f262c08` | latent build loss — same attribute |
[...]

Issue #228, opened 2026-09-02
waterui-canvas: Path::arc_to is not HTML5 arcTo — broken angle, bisector, and subpath


lexoliu:
`waterui_canvas::Path::arc_to` documents itself as "equivalent to HTML5 Canvas `arcTo()`",
and is not. It has no callers and no tests anywhere in the workspace, so nothing has ever
depended on it; the first caller gets a broken shape.

Four defects, in `components/visual/canvas/src/path.rs:128`:

**1. The arc is appended with its leading `MoveTo`.**

```rust
let arc_path = arc.to_path(0.1);
for el in arc_path.elements() {
    self.inner.push(*el);
}
```

`kurbo::Arc::to_path` returns a path that begins with a `MoveTo` to the arc's start point.
Pushing that into the current path lifts the pen and starts a new contour, so every corner
breaks the outline into disconnected fragments and a later `close()` closes the wrong one.

**2. The angle is the deflection, not the interior angle.**

```rust
let v0 = p1 - p0;              // incoming direction
let v1 = p2 - p1;              // outgoing direction
let angle = v0_norm.dot(v1_norm).acos();
```

A corner fillet is defined by the interior angle between the two segments *as seen from the
corner* — between `p1 -> p0` and `p1 -> p2`. What is computed is its supplement, so the
tangent distance `r / tan(angle/2)` comes out as `r * tan(theta/2)` instead. A right angle
happens to survive (both give `r`); a shallow turn does not. For a 10-degree deflection the
tangent distance should be about `0.087 * r` and comes out about `11.4 * r`, which is what
turns a routed polyline into strays that shoot off the diagram.

**3. The bisector points the wrong way.**

```rust
let bisector = (v0_norm + v1_norm).normalize();
```

The arc's centre lies along the bisector of the two vectors *from* the corner, i.e.
`normalize((p0 - p1).normalize() + (p2 - p1).normalize())`. Adding the travel directions
instead puts the centre 90 degrees off even for a right angle.

**4. Degenerate cases silently draw nothing.**

When the points are collinear, or a segment has zero length, the function returns without
touching the path. HTML5 `arcTo` draws a straight line to `point1` in exactly those cases;
returning leaves a gap.

Reproduced by drawing a rounded rectangle (`Go` below is `A([Go])`, a stadium) and a routed
three-point edge — both come out as loose fragments:

- stadium rendered as four disconnected strokes rather than a closed pill
[...]

PR #438, opened 2026-09-08
fix(mermaid): pin waterui-canvas to the restored HTML5 arc_to

### LINKED ISSUE #418: mermaid: a flowchart with a cycle draws its nodes off-canvas and its edges detached from them
`cargo nextest run -p waterui-mermaid -E 'test(export_flowchart_images_for_visual_review)' --no-capture` exports two flowcharts. The `subgraphs` case (`flowchart LR`, a DAG) is correct: boxes, joined edges, arrowheads at the node borders. The `decision` case is not.

```
flowchart TD
    A[Start] --> B{Ready?}
    B -->|yes| C([Go])
    B -->|no| D[(Wait)]
    D --> A
```

In an 800x600 viewport the whole diagram is crushed into the top-left quadrant, and:

- the `Ready?` diamond is cut off by the left edge of the canvas — its left vertex is at a negative x;
- no edge meets the node it belongs to. The `Start` connector curls away to the right of the box instead of leaving its bottom; `Ready? -> Go` and `Ready? -> Wait` are floating stubs that touch neither end; `Wait -> Start` runs down and to the right, off into empty space, away from `Start` which is at the top;
- the `yes`/`no` edge labels sit in open space, far from the edges they name;
- nothing is centred: the bottom half and the right half of the canvas are empty.

The two cases differ in direction (`TD` vs `LR`) and in that `decision` has a cycle (`D --> A`), which is the likely reason the layout emits coordinates the drawing side does not expect. Node geometry and edge geometry clearly disagree about the origin, so the two are being placed in different coordinate spaces rather than one being wrong on its own.

This is a rendering defect, not a measurement one: it is identical before and after the font-collection change in #409/#417.

lexoliu:
A `flowchart TD` with a cycle draws its edges as fragments: the `Start` connector curls off to the side instead of leaving the box, the two decision branches touch neither end, and `Wait --> Start` runs down and to the right, off the diagram entirely.

The layout was never at fault. Every node box and every routed polyline `merman` hands over is inside the diagram's bounds, and each polyline already ends on the border of the node it belongs to. The painting came apart.

`draw::route` rounds each interior corner of a routed polyline with `Path::arc_to`, the platform's corner-fillet primitive, and that primitive is broken in the published `waterui-canvas` 0.1.0 three ways at once:

- it takes the deflection between the travel directions for the interior angle at the corner, so the tangent distance comes out as `r * tan(θ/2)` instead of `r / tan(θ/2)`;
- it places the arc's centre on the bisector of the travel directions rather than of the two vectors leaving the corner, ninety degrees away;
- it splices the `MoveTo` that `kurbo::Arc::to_path` opens with into the contour, lifting the pen mid-edge.

A right angle survives the first two by coincidence, which is exactly why the `LR` diagram of straight segments always looked right while a forty-degree turn shoots the path far past the node it was heading for.

**This is a regression, not a latent gap.** The monorepo fixed `arc_to` in `ef11533` (#228) on 2 September. The standalone repository had already forked without it, so when `cba85e1c` dropped the in-tree copy two days later and took `waterui-canvas` 0.1.0 from crates.io, the fix went with it. That failure mode is now audited across every split-out crate in #431 — `arc_to` was the only functional loss — and the pin's removal is tracked in #432.

Pins the crate to water-rs/canvas#5 (merged), which restores the implementation unchanged and adds `Path::bounding_box` so a consumer can assert what it drew.

Three tests in `draw` cover the geometry the pictures were showing: every node outline stays on the canvas, every edge path stays inside the box its own polyline spans and on the canvas, and every edge ends on a node. The middle one fails against the published `arc_to`.

Verified by eye on the re-exported `decision` flowchart: every edge meets its node, `Wait --> Start` routes up the right side into the top of `Start`, the diamond is fully on canvas, and `Go` is a closed pill. 28/28 mermaid tests, clippy clean.

Fixes #418

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,9 @@ name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
+dependencies = [
+ "derive_arbitrary",
+]
 
 [[package]]
 name = "arboard"
@@ -2607,6 +2610,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive_arbitrary"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e567bd82dcff979e4b03460c307b3cdc9e96fde3d73bed1496d2bc75d9dd62a"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
+]
+
+[[package]]
 name = "derive_more"
 version = "2.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


lexoliu:
Force-pushed `da88a7330`: the lockfile in the previous head listed `waterui-canvas` twice — two byte-identical `[[package]]` blocks for the same git revision — which `cargo` accepts for building but refuses to parse for `cargo update`:

```
error: failed to parse lock file at: Cargo.lock
Caused by:
  package `waterui-canvas` is specified twice in the lockfile
```

That would have broken the weekly dependency check, which is exactly the kind of thing that goes unnoticed because the build stays green. Found by running `cargo update` against this branch while working on #405. Duplicate removed, `cargo metadata` parses, `cargo check -p waterui-mermaid` clean.

PR #445, opened 2026-09-08
build(deps): take the outstanding independent dependency releases

lexoliu:
Refs #405 — the routine half of it.

Ten crates the workspace pins by version had newer releases whose upgrades are
independent of one another and of the renderer stack:

| Crate | From | To |
| --- | --- | --- |
| `quick-xml` | 0.41 | 0.42 |
| `lru` | 0.16 | 0.18 |
| `geo` | 0.32 | 0.33 |
| `glow` | 0.17 | 0.18 |
| `getrandom` | 0.2 | 0.4 |
| `skrifa` | 0.42.1 | 0.46.2 |
| `usvg` | 0.46.0 | 0.48.1 |
| `unicode-segmentation` | 1.12 | 1.13 |
| `accesskit` + `accesskit_winit` | 0.24 / 0.33.2 | 0.25 / 0.34.0 |

One source change came with them: `accesskit` 0.25 renames `Tree` to `TreeInfo`
and leaves the old name as a deprecated alias, so the dew and hydrolysis
accessibility trees name the new type rather than compiling against a
deprecation — this workspace treats a deprecation warning as an error, so
"it still compiles" was not an option.

Two are deliberately not here:

- **the wgpu / vello / vello_cpu / zenwave stack**, which is one coordinated move
  behind a forked wgpu (the Mali negative-viewport patch) rather than a set of
  routine bumps, and belongs in its own change with its own device verification;
- **`base64` 0.22 → 0.23**, which also lives in the `kit` submodule, so it needs a
  kit change and a pin bump rather than a line in this manifest.

Verified locally on the full workspace: `cargo clippy --workspace --all-targets -D warnings`
clean, and `cargo nextest run --workspace` at 1738 passed / 1 failed — the single
failure being `fallback_release_backend_refs_match_workspace_backend_refs`,
which was reading the workspace's *checked-out* Android submodule (moved for an
unrelated branch) against the pin, and passes at the pinned commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,46 +20,46 @@ checksum = "366ffbaa4442f4684d91e2cd7c5ea7c4ed8add41959a31447066e279e432b618"
 
 [[package]]
 name = "accesskit"
-version = "0.24.1"
+version = "0.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3b7f7f85a7e5f68090000ed7622545829afd484d210358702ae4cb97dd0c320"
+checksum = "438a7081a65b95c668db56591a4fef9bc5dad275f448bd6223fc6949d823db38"
 dependencies = [
  "uuid",
 ]
 
 [[package]]
 name = "accesskit_atspi_common"
-version = "0.19.1"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "023da0e5097f46df7092d5280b02efb9bbf8d93298daeced42652463e357d636"
+checksum = "c9d47ad644916f6cb7e432a5ca0dbd7cc78281a9772881057bb72d4aadb93257"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
  "atspi-common",
- "phf",
+ "phf 0.14.0",
  "serde",
  "zvariant",
 ]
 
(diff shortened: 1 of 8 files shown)


lexoliu:
`glow` comes back out of the table. It is not a routine pin: the GTK backend hands a `glow::NativeFramebuffer` straight to `wgpu::hal::gles::TextureInner::ExternalNativeFramebuffer`, and `wgpu-hal` links glow **0.17**, so taking 0.18 put two copies of the crate in the graph and turned one name into two types.

```
error[E0308]: mismatched types
  --> backends/gtk/src/components/graphics/gpu_surface.rs:925:86
    |
925 |   inner: wgpu::hal::gles::TextureInner::ExternalNativeFramebuffer { inner: attachment },
    |                                                                           ^^^^^^^^^^
    |     expected `glow::native::NativeFramebuffer`, found `glow::NativeFramebuffer`
note: there are multiple different versions of crate `glow` in the dependency graph
```

That is the same reason the wgpu / vello / vello_cpu / zenwave stack is already excluded here — glow belongs to it. The pin now carries a comment saying so, so the next sweep does not raise it on its own again.

Verified locally against GTK 4.20.3: `cargo check -p waterui-gtk --all-targets` is clean, and the lock resolves a single `glow 0.17.0` that `wgpu-hal` shares.

PR #449, opened 2026-09-08
feat(waterui): forward the markdown-math feature through the facade

### LINKED ISSUE #446: The markdown-math feature is unreachable from an application
`waterui-internal` defines `markdown-math`:

```toml
# Mathematical formulas in Markdown: `$…$` and `$$…$$` are parsed and typeset
# through `waterui-math`. Off by default, and deliberately so — …
markdown-math = ["flow-markdown", "dep:waterui-math"]
```

The public `waterui` facade forwards `flow-markdown`, `highlight`, `video`,
`webview` and the rest to it — but not this one. Every application depends on
`waterui`, so there is no way for one to turn the feature on:

```
package `flow-markdown-example` depends on `waterui` with feature `markdown-math`
but `waterui` does not have that feature.
help: available features: all, assets, default, dynamic_linking, flow-markdown,
gpu, highlight, inspector, inspector-signals, locale-datetime, locale-list,
media, navigation-restoration, snackbar, std, video, video-gpu, webview
```

So math in Markdown ships in the tree, is documented as a knob an application
turns on, is tested inside the workspace — and cannot be reached by any consumer
of the framework. `$x^2$` in an application's Markdown stays a dollar sign
whatever it asks for.

The fix is the missing forwarding line, matching every neighbour:

```toml
markdown-math = ["waterui-internal/markdown-math", "waterui-dylib?/markdown-math"]
```

Whether it also joins the facade's `all` is the real question, and the reason
this is an issue rather than a one-line drive-by. `waterui-internal`'s own `all`
includes it; the facade's `all` is what CI builds on every platform, and the
feature pulls a layout engine, an OpenType `MATH` table reader, a LaTeX parser
and MathCAT's speech rules — about 2.7 MiB — plus a bundled math font for the
asset pipeline. `--features all` is also what the Windows lane uses in place of
`--all-features`. Adding it there changes what every full-feature CI leg builds
and downloads.

Found while verifying the GPU-surface accessibility label (#394): a formula is
the only content in the tree that gives its surface a default label, and there
is no way to put one on screen from an application.

lexoliu:
`waterui-internal` and the Markdown component have carried `markdown-math` since math landed, but neither the `waterui` facade nor `waterui-dylib` re-exported it. An application depending on `waterui` therefore had no way to turn it on — `waterui = { workspace = true, features = ["markdown-math"] }` failed to resolve, and `$…$` in Markdown rendered as literal text.

The feature stays **outside `all`** on purpose, and the manifest says why: it pulls a layout engine, an OpenType `MATH` table reader, a LaTeX parser, MathCAT's speech rules and a bundled math font — about 2.7 MiB — and `all` is what every full-feature CI leg builds on every platform.

Verified by building and running the `flow_markdown` example on macOS with the feature enabled: `$r = \\frac{e}{b}$` renders as a formula instead of literal text.

Fixes #446

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -339,6 +339,11 @@ navigation-restoration = [
     "waterui-dylib?/navigation-restoration",
 ]
 flow-markdown = ["waterui-internal/flow-markdown", "waterui-dylib?/flow-markdown"]
+# Mathematical formulas in Markdown. Deliberately outside `all`: it brings a
+# layout engine, an OpenType `MATH` table reader, a LaTeX parser and MathCAT's
+# speech rules — about 2.7 MiB — and a bundled math font for the asset pipeline,
+# and `all` is what every full-feature CI leg builds on every platform.
+markdown-math = ["waterui-internal/markdown-math", "waterui-dylib?/markdown-math"]
 std = ["waterui-internal/std", "waterui-dylib?/std"]
 dynamic_linking = ["dep:waterui-dylib"]
 # Every feature the framework offers an application. `dynamic_linking` is
--- a/deny.toml
+++ b/deny.toml
@@ -17,6 +17,13 @@ allow = [
     "BSD-2-Clause",
     "BSD-3-Clause",
     "BSL-1.0",
+    # MathCAT's build script unpacks its speech-rule bundle with `zip`, which
+    # brings `libbz2-rs-sys`. The bzip2 licence is BSD-style and permissive —
+    # redistribution in source and binary form, keep the notice, do not
+    # misrepresent the origin — so it sits in the same class as `Zlib` and
+    # `0BSD` above. It reaches no shipped artifact either way: the crate is a
+    # build dependency and its code never enters an application binary.
+    "bzip2-1.0.6",
     "CC0-1.0",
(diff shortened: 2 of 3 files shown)


lexoliu:
Rebased onto `dev` (post-#448) and added the `bzip2-1.0.6` allowance that Hygiene asked for.

Making `markdown-math` reachable from the facade puts MathCAT's build script inside the `cargo deny` all-features graph, and with it `libbz2-rs-sys` — `zip` uses bzip2 to unpack MathCAT's speech-rule bundle at build time. Reproduced locally against post-#448 `dev` with a stable lock, so it is the feature and not a stale-lock re-resolve:

```
error[rejected]: failed to satisfy license requirements
  libbz2-rs-sys v0.2.5  → bzip2 v0.6.1 → zip v6.0.0 → (build) mathcat v0.7.5
```

The bzip2 licence is BSD-style and permissive — redistribute in source or binary form, keep the notice, do not misrepresent the origin — so it sits in the same class as `Zlib` and `0BSD` already on the list, and the crate is a build dependency whose code never enters an application binary. `cargo deny check licenses bans` is green with the entry.

Issue #460, opened 2026-09-09
Promote only a fully tested revision to the nightly distribution channel


lexoliu:
Nightly is a consumable framework version, not merely a scheduled test run.

Implement promotion of the exact WaterUI commit tested by the full integration suite, including its pinned submodule commits. Publish an immutable identity for that successful revision and update the discoverable nightly channel only after every required check succeeds. Failed, cancelled, incomplete or superseded runs must not replace the last successful nightly. Keep dev (compile-checked integration branch), nightly (test-certified revision), and stable (crates.io releases) distinct. Stable publishing remains owned by release-plz and must not be triggered by nightly promotion.

Acceptance: exercise successful and failed promotion paths; prove consumers resolve the tested commit and matching backends/submodules, including when dev advances during the run. The implementation PR targets dev.

lexoliu:
Distribution design: use immutable nightly-<date>-<commit> tags with GitHub prereleases, not a moving nightly branch. Promote the exact tested dev commit only after all required integration checks pass; its gitlinks pin the component/backend revisions. Failed or cancelled runs publish no candidate, and older runs must not make the channel regress. CLI resolves the latest eligible nightly to an exact revision and persists it; ordinary builds do not upgrade it. No certified nightly means an explicit error, not a fallback to dev. Nightly promotion must not publish to crates.io or invoke stable release-plz automation.

lexoliu:
The independently extracted drawing components will be versioned crate dependencies, not new submodules. A nightly certifies the framework commit together with the resolved dependency set tested by the suite; CLI consumers must preserve that dependency lock instead of re-resolving newer components. Existing native-backend gitlinks remain part of the fixed framework revision. The immutable-tag design is unchanged.

Issue #459, opened 2026-09-09
Cargo.lock again records a second canvas source after the dependency update


lexoliu:
The committed dev tree at 686eef60ba2359c4399f5aaf4c7bf76638b321bc fails `cargo metadata --locked --offline --format-version 1`: Cargo must remove the registry copy of waterui-canvas and resolve waterui-chart through the existing canvas patch. This is a pre-existing recurrence of #447 after the dependency update, exposed while verifying #458. The canonical working tree already contains exactly that lockfile correction. Preserve and integrate that correction, without updating unrelated dependencies, and verify locked metadata resolution succeeds.

lexoliu:
The existing canonical Cargo.lock correction has been applied unchanged in the isolated workspace. `cargo metadata --locked --offline --format-version 1` now succeeds; it failed before the correction. No dependency upgrade was performed. This preserves the local work rather than regenerating a different lockfile. Awaiting review before committing/publishing the dedicated fix.

PR #463, opened 2026-09-09
ci: reserve the dev gate for cached compilation checks

### LINKED ISSUE #458: ci: a push to dev must finish its gate in under 10 minutes (measured: 45 min mean, compile-bound, zero target caching)
Target (2026-09-08): **a push to `dev` must finish CI in under 10 minutes.**

## Measured baseline

Post-#379, the slim gate on dev pushes averages **44.6 min** over 11 completed runs (median 38.9, p25 30.2, p75 43.1, max 128.6), 2026-09-07T05:34Z → 2026-09-08T09:44Z. 30 of 41 pushes were cancelled mid-flight by the concurrency group — merge-train supersession, not failures.

Critical path of the last green run (34211677737, #445 merge):

| Job | Duration |
|---|---|
| `Test / ubuntu-latest` | **40 min — the gate** |
| `Test / lint` | 22 min |
| `Test / examples (ubuntu-latest)` | 21 min |
| `Test / wasm32` | 7.5 min |
| Hygiene / Typos / Changes | seconds |
| FFI Header / all-features / Android FFI / macos-suites | path-filtered off ✅ |

Inside `ubuntu-latest`, one step is the whole story: `cargo nextest run --workspace --exclude '*-example' --profile ci` ran **38 min** (doctests 1.3 min, everything else seconds).

## Root cause

`cache-targets: false` on the test legs: the 10 GB Actions cache budget forced the compromise where **no target directory is cached**, so every dev push cold-compiles the whole workspace test profile. The 38 minutes is compile-bound, not test-bound. (#328/#362 context.)

## Option space (a route is a design choice, not yet decided)

1. **Shard the nextest step** into N parallel jobs — the `run_examples_shard.sh` pattern already exists in android-backend. 4 shards ≈ 10 min each. Mechanical; does not touch the compile cost itself.
2. **Bring compilation caching back for the gate profile** — reserve a dedicated cache slice for the dev-push profile (today's compromise caches no target at all), or move to sccache-style remote storage to escape the 10 GB Actions cap entirely.
3. **Incremental testing** — on a dev push, run nextest only over the package subgraph the merge's diff touches relative to the last green dev run.
4. **Question the run itself** — a merge commit's tree is the same tree the PR head already ran green. The dev-push run only earns its keep against semantic conflicts between two individually-green PRs (what merge queues exist for). Cutting it to a smoke check is the fastest route but requires an explicit change to the gate policy.

Note the interplay with the 73% cancel rate: under a merge train, the effective green signal often arrives a run later than the push that needed it, so the perceived latency is worse than the mean suggests.

lexoliu:
## Summary
- Replace the pre-existing test-heavy dev push gate with format and default-feature workspace clippy.
- Preserve every existing test command in the full nightly invocation, including examples, doctests, wasm and skill snippets.
- Give the frequent check-only profile priority over nightly test archives in the Actions cache budget.

Fixes #458

Depends on the locked canvas source correction in #459; that prerequisite is a separate PR and should land first.

#### Test plan
- [x] actionlint on the five affected workflows.
- [x] Four policy checks: dev routing, fmt/clippy-only fast path, unchanged test commands, and lint cache ownership.
- [x] Locked offline dependency resolution and `git diff --check`.
- [ ] Measure cold and warm Actions runs; a timeout is not evidence that the ten-minute target is met.
- [ ] Verify the full nightly matrix remains runnable.

Repository extraction and distributed component CI remain in #186/#194. Certified nightly distribution and CLI channel selection are tracked in #460/#461.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ permissions:
   # dorny/paths-filter lists a pull request's changed files.
   pull-requests: read
 
-# The per-change gate (#379): what a pull request or a push to `dev` runs.
+# The pull-request gate (#379); dev pushes use the lint-only dev.yml (#458).
 # It is deliberately the default-features shape on Linux plus the checks a
 # change can break, chosen per change by the `changes` job below. The full
 # matrix — every clippy shape, macOS and Windows, the all-features runs, the
@@ -17,7 +17,7 @@ on:
   # Integration branches only: a branch that also has a PR would otherwise run
   # the whole gate twice for one commit.
   push:
-    branches: [main, dev]
+    branches: [main]
   pull_request:
     branches: ["**"]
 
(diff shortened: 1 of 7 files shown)


lexoliu:
First Actions measurement: the new fast lint job passed in 9m26s on a cold dedicated cache (CI run 34300073350, job 102304932640). Format took 4s; the complete job includes checkout, toolchain setup and native dependencies. This measures the reusable job that dev.yml calls, not a completed dev-push workflow: the PR still runs its separate ABI/security/benchmark checks, and the dev event plus warm-cache behavior remain to be verified after integration. No lint scope was narrowed and no warnings were suppressed.

Issue #468, opened 2026-09-09
Docs-only pull requests trigger Rust CI because paths-filter every is applied to patterns


lexoliu:
Pre-existing defect reproduced by #467: its only changed file is AGENTS.md, yet CI run 34301997692 schedules lint, workspace tests, examples and wasm.

The prose filter in .github/workflows/ci.yml uses `predicate-quantifier: every` with four positive patterns (`**/*.md`, `docs/**`, `LICENSE*`, `.github/ISSUE_TEMPLATE/**`). In dorny/paths-filter, every means a single file must match every pattern; it does not mean every changed file matches one of the documentation patterns. Consequently docs_only is false and the code output is true even for AGENTS.md alone.

Detect whether any changed file lies outside all documentation patterns using every with negated patterns, and use that result directly for the code output. Preserve the explicit workflow override and mixed documentation/code behavior.

Acceptance: regression cases cover root/nested Markdown, docs, licenses, issue templates, ordinary source, mixed changes, and workflow/action files; docs-only PRs do not launch Rust compilation; real code or workflow changes still do. Use a dedicated PR targeting dev.

lexoliu:
Verified against the real docs-only PR #472: Changes succeeded; Hygiene, FFI Header, Android FFI, Test and Test Report were skipped. The final #470 diff changes only the existing path-filter configuration; no CI self-test job or permanent policy suite was added.

PR #472, opened 2026-09-09
docs: favor coherent redesign and selective durable tests

### LINKED ISSUE #469: Clarify design integrity and selective regression test guidance
AGENTS.md currently leaves overengineering ambiguous and can be read as favoring small patches over necessary redesign. Prefer coherent structural fixes, explicit contracts instead of speculative defensive branches, and selective durable tests. Keep one-off diagnostic checks out of the repository and consider cumulative suite cost rather than adding a regression test for every concern. Update the guidance without weakening meaningful product coverage or input validation at real trust boundaries.

lexoliu:
## Summary
- Prefer coherent structural redesign over accumulating patches or minimizing the diff.
- Avoid speculative defensive programming while preserving real trust-boundary validation.
- Keep one-off checks temporary and add durable regression tests selectively, accounting for redundancy and cumulative suite cost.

Fixes #469

#### Test plan
- [x] typos AGENTS.md and git diff --check.
- [x] Only AGENTS.md differs from current dev.
- [ ] Confirm the docs-only CI run skips Rust jobs, validating #468 on an actual pull request.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,15 @@ These are the target architecture and acceptance criteria for repository changes
 
 ## Engagement Rules
 
-DO NOT be over-engineer or write defensive code. If you encounter a problem, ask user for solution with your own idea, do not say "Let's have a simpler approach". You are expected to face the real problem and make code clean, reusable and elegant. Never take a workaround.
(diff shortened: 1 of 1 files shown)


lexoliu:
Superseded by #475 with the same tree and a verified signed commit. The docs-only CI result here remains the path-filter verification evidence.

PR #473, opened 2026-09-09
feat(waterui): re-export optional standalone components

### LINKED ISSUE #464: Expose independently versioned components through the waterui facade
Keep waterui as the unified facade, including independently maintained components such as chart. Repository extraction changes implementation ownership and CI, not the unified application API.

Add direct, feature-gated re-exports of the standalone component crates using their existing workspace dependency declarations. Avoid wrapper types or duplicate implementations. Applications should be able to use the facade path or depend directly on the component crate. Keep heavyweight dependencies optional and verify that disabling a feature removes its dependency graph.

Maintain the normal dependency direction: waterui facade -> component -> public graphics/core APIs. Component implementations must not normally depend on the facade that re-exports them. Check that facade and direct imports share the same types and compatible dependency resolution, including the dynamic-linking path. Backend re-exports are not implicit in this change: backend-to-facade dependencies must be addressed before adding a reverse edge.

Acceptance: chart is reachable through the facade; the selected standalone components have coherent feature forwarding; compile tests exercise both facade and direct imports, with default features disabled as well as enabled. Use a dedicated implementation PR targeting dev.

lexoliu:
## Summary
- Expose canvas, chart, barcode and particle directly through feature-gated facade modules.
- Forward the features through waterui-internal and the dynamic-link anchor without wrapper types or reverse dependencies.
- Update the existing authoring guidance; retain independent crate imports and ownership.

Fixes #464

#### Test plan
- [x] Compile each component independently with default features disabled.
- [x] Compile the combined features with defaults disabled, enabled and with dynamic linking.
- [x] Compile direct-import-to-facade type assignments for all seven shapes.
- [x] Verify the lean normal dependency graph omits all four optional components.
- [x] Scoped clippy with -D warnings, rustfmt check and typos.
- [x] Keep the temporary verification harness outside tracked source.


lexoliu:
Superseded by #476 with the same tree and a verified signed commit.

Issue #533, opened 2026-09-11
Foundation 0.4 release: extracted registry crates pin graphics ^0.3, so the release commit does not build


lexoliu:
Release PR #532 bumps `waterui-graphics` and `waterui-text` to 0.4.0 (cargo-semver-checks reports API breaking changes). Every crate the workspace consumes from crates.io — `waterui-barcode`, `waterui-canvas`, `waterui-chart`, `waterui-image`, `waterui-map-gpu`, `waterui-particle`, `waterui-video-gpu`, `waterui-visualizer` — requires `waterui-graphics ^0.3` (chart and video-gpu also `waterui-text ^0.3`). The `[patch.crates-io]` entries that map those requirements onto the workspace copies stop matching at 0.4.0, cargo resolves the registry 0.3.0 copies beside the path 0.4.0 ones, and every job on #532 fails with

```
error[E0277]: the trait bound `waterui_core::Str: From<waterui_str::Str>` is not satisfied
  --> waterui-text-0.3.0/src/text.rs:246:41
```

Two structural facts shape the fix:

- `waterui-media` depends on `waterui-image` unconditionally and sits in the umbrella's default `media` feature, so `waterui-media`, `waterui-internal`, `waterui-dylib`, `waterui` and `waterui-testing` cannot pass `cargo publish` verification until `waterui-image` re-releases against graphics 0.4. `hydrolysis` and `hydrolysis-m3` only use `waterui-canvas` in dev-dependencies and can publish first.
- The extracted repositories dev-depend on the umbrella and on `hydrolysis-m3` for their tests, which forms a cross-repository dev-dependency cycle. `water-rs/image`'s lockfile already resolves two copies of `waterui-image` (the registry one through umbrella → media, plus itself). Each extracted repository needs `[patch.crates-io] waterui-<self> = { path = "." }` so its tests always link the local crate.

Release sequence:

1. Each extracted repository gets a `chore/foundation-0.4` branch moving its foundation requirements to graphics 0.4 / text 0.4 / testing 0.4 (registry requirements, replacing the `git = …, branch = "dev"` pins where present) plus whatever the API changes require. `water-rs/canvas` branches from its released `main` (0.1.1, never merged back into `dev`); `water-rs/visualizer` sets its version to the registry's 0.3.0 so release-plz bumps from the published baseline.
[...]

PR #534, opened 2026-09-11
Merge the 0.4 foundation release back into dev

lexoliu:
Brings the release-plz version bumps and changelogs from #532, plus the temporary `[patch.crates-io]` entries that resolve the extracted crates from their `chore/foundation-0.4` branches (#533), back into `dev` so the follow-up that drops those entries can be built on top of the released versions.


lexoliu:
Superseded by a dev-based merge branch: main and dev conflicted on Cargo.toml/Cargo.lock (uuid bump on dev vs the core version bump on main).

PR #537, opened 2026-09-11
Release: dev → main (foundation 0.4 registry pins)

lexoliu:
Takes the eight extracted crates from crates.io at their foundation 0.4 releases (barcode 0.2.0, canvas 0.1.2, chart 0.2.0, image 0.4.0, map-gpu 0.2.0, particle 0.1.1, video-gpu 0.3.1, visualizer 0.4.0) and removes the git patch entries that stood in for them, so `main` can publish the crates that the release run has not reached yet.

`waterui-mermaid` is marked `publish = false`: crates.io refuses its git-only merman dependencies, and the run aborting there is what left browser-cef, browser-wpe, chromium, cli, dew, ffi, gtk, the icon sets, inspector-app and preview unpublished. Only the markdown example depends on it.

Refs #533


lexoliu:
`Local API vs published baseline` is red, and not because of this change: the `current` side built in 242 s and the failure is in the **baseline** build, where the published `waterui` 0.4.0 pulls the pre-0.4 extracted crates (chart 0.1.0, shape 0.1.1) and so carries two copies of `waterui-graphics`. That is the exact condition #533 clears, and merging this is what clears it. Recorded as #538.

PR #541, opened 2026-09-11
ci(release): give the CLI binary legs the tools they build with

### LINKED ISSUE #540: The CLI release builds no binaries: seven of eight legs fail
The `waterui-cli-v0.2.0` release (run 34583623908) built one target out of
eight. `Upload Release Assets` needs every build leg, so it was skipped and
the release page carries no `water` binary at all.

Seven legs, three causes in this repository — all pre-existing, none
introduced by the release itself:

**Linux builds have no native dependencies.** `Build x86_64-unknown-linux-gnu`
stops at `cros-libva v0.0.12`: "libva is not found in system". Every other
Linux job in this repository installs that set through
`.github/actions/setup-linux-deps`; `build-cli.yml` never adopted it.

**aarch64-linux is cross-compiled.** The same leg on `ubuntu-latest` adds
"pkg-config has not been configured to support cross-compilation", so even with
the headers present a cross build cannot find them. GitHub now offers
`ubuntu-24.04-arm`, which the WPE legs of this same workflow already use.

**Windows images ship no `dxc`.** Both `x86_64-pc-windows-msvc` and
`aarch64-pc-windows-msvc` die in `filtrate`'s build script: "DXIL compilation
failed for src/shaders/shared\blit.wgsl: unable to execute dxc: program not
found". `windows.yml` and `bench.yml` install it from
`tracel-ai/github-actions/install-dxc`; this workflow does not.

**`refresh-apt-index` assumes Ubuntu 24.04.** It passes
`-o Dir::Etc::SourceList=/etc/apt/sources.list.d/ubuntu.sources`, which exists
only in 24.04's deb822 layout. The two WPE runtime legs run on `ubuntu-22.04`
and `ubuntu-22.04-arm`, where apt resolves an empty index instead of failing,
reports `E: Unable to locate package` for the whole list, and
`setup-linux-deps`'s verification step aborts the job.

The macOS x86_64 leg failed for a fourth reason that lives in another
repository: `Float16` is unavailable when targeting x86_64 macOS, which
water-rs/waterkit#51 covers and water-rs/waterkit#52 fixes.

lexoliu:
The 0.2.0 CLI release built one binary out of eight and skipped
`Upload Release Assets`, so the tag carries no download. Each failure is a
tool the job never installed; all three gaps are pre-existing.

* **Linux legs** now run `.github/actions/setup-linux-deps`, the same action
  every other Linux job uses, so `cros-libva` finds VA-API's headers.
* **aarch64-linux** builds natively on `ubuntu-24.04-arm` instead of
  cross-compiling from x86_64, where `pkg-config` refuses to resolve anything.
  The WPE legs of this workflow already run on arm runners.
* **Windows legs** install `dxc` from the same pinned
  `tracel-ai/github-actions/install-dxc` that `windows.yml` and `bench.yml`
  use, so `filtrate` can compile its shaders to DXIL.
* **`refresh-apt-index`** picks the source file the runner actually has:
  24.04's deb822 `ubuntu.sources` when it exists, 22.04's one-line
  `sources.list` otherwise, and an explicit error when neither does. Naming
  the missing file resolved an empty index, which is why both WPE runtime
  legs reported every package as unlocatable.

The remaining leg, `x86_64-apple-darwin`, fails on `Float16` in the codec
kit's Swift sources; water-rs/waterkit#52 fixes that and lands through the
`kit` submodule pin.

Fixes #540

### DIFF
--- a/.github/actions/refresh-apt-index/action.yml
+++ b/.github/actions/refresh-apt-index/action.yml
@@ -13,9 +13,21 @@ runs:
   steps:
     - name: Refresh the Ubuntu package index
       shell: bash
-      # Ubuntu 24.04 declares its archive in deb822 form at this path; apt
-      # picks the parser from the `.sources` extension.
-      run: >
-        sudo apt-get update
-        -o Dir::Etc::SourceList=/etc/apt/sources.list.d/ubuntu.sources
-        -o Dir::Etc::SourceParts=/dev/null
+      # Ubuntu 24.04 declares its archive in deb822 form at
+      # `sources.list.d/ubuntu.sources`; 22.04 still uses the one-line
+      # `sources.list`, and the runners here are both. Naming the file that is
+      # not there resolves an empty index instead of failing, so apt then
+      # reports every package as unlocatable and the install step aborts —
+      # which is how the WPE runtime legs of the CLI release died.
+      run: |
+        if [ -f /etc/apt/sources.list.d/ubuntu.sources ]; then
+          list=/etc/apt/sources.list.d/ubuntu.sources
+        elif [ -s /etc/apt/sources.list ]; then
+          list=/etc/apt/sources.list
+        else
+          echo "::error::no Ubuntu archive source file on this runner"
+          exit 1
+        fi
+        sudo apt-get update \
+          -o Dir::Etc::SourceList="$list" \
+          -o Dir::Etc::SourceParts=/dev/null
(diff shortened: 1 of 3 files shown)


lexoliu:
Added the fourth cause: the `kit` submodule now points at water-rs/waterkit#52, merged, which is what makes the codec kit's Swift sources compile for x86_64 macOS. All four legs of #540 are now addressed in this branch.

Issue #538, opened 2026-09-11
Release preflight cannot build the published 0.4.0 baseline


lexoliu:
`release-preflight.yml`'s "Local API vs published baseline" job fails on pull requests to `main` while it builds the **baseline**, not the change. Run 34580457192 on #537:

```
Building waterui v0.4.0 (current)   ->  Built [242.551s]
Building waterui v0.4.0 (baseline)  ->  error[E0277]: the trait bound
    `waterui_graphics::Color: Into<waterui_graphics::color::Color>` is not satisfied
    --> registry/src/.../waterui-chart-0.1.0/src/legend.rs:141:56
    --> registry/src/.../waterui-shape-0.1.1/src/lib.rs:1254:31
```

The baseline is the published `waterui` 0.4.0, which requires the extracted crates at their pre-foundation-0.4 versions (chart 0.1.0, image 0.3.0, and so on). Those require `waterui-graphics ^0.3`, so the baseline's own graph carries graphics 0.3 beside 0.4 and does not compile — the condition #533 exists to clear. Nothing outside the registry can fix that graph; it is a property of what is already published.

It resolves itself with the next release: once `waterui` publishes a version whose extracted requirements are the 0.4 releases, the baseline moves to that version and builds. This issue is here so the recovery is checked rather than assumed — if the job is still red after the next publish, the cause is a different one.

lexoliu:
Fixed by the 0.4.1 publish.

The check could not build the baseline because `waterui` 0.4.0 on crates.io pinned `waterui-graphics` 0.3.0 alongside a `waterui-chart` 0.1.0 that had already moved to the corrected `Color` conversion, so rustdoc for the baseline died on `waterui_graphics::Color: Into<waterui_graphics::color::Color>`. That mismatch existed only in the published 0.4.0 line; 0.4.1 ships the whole set together.

Verified rather than assumed: dispatched `release-preflight.yml` against `dev` now that 0.4.1 is the newest unyanked version — [run 34632560877](https://github.com/water-rs/waterui/actions/runs/34632560877), `Local API vs published baseline` **success**.

PR #545, opened 2026-09-11
Merge dev: CLI release binaries, avatar cover-crop, CI gate documentation

### LINKED ISSUE #427: waterui-image has no content mode, so a non-square picture is stretched instead of cover-cropped
Found while landing the Avatar component (#93). An avatar drawn from a non-square picture is stretched to fill its circle rather than cover-cropped, which is what every avatar in the wild does and what the shape implies.

`ReactiveImage` is either intrinsic-sized or `resizable()`, and `AspectRatio` cannot bridge the gap because a resizable image reports no intrinsic aspect ratio — there is nothing for the aspect modifier to preserve. So there is no way, from application code, to say "fill this frame, preserve the picture's aspect, crop the overflow".

The gap is in `waterui-image`, which is consumed from crates.io and developed in its own repository, so the fix does not land in this tree. Recording it here so the limitation is written down where the components that hit it live: Avatar is the first, and any component that puts a picture in a fixed frame — a card header, a thumbnail grid, a cover image — will hit the same wall.

What is missing is a content mode on the image itself (fit / fill / cover, in the usual naming), decided where the decoded size is known, rather than anything an outer modifier can express.

### LINKED ISSUE #540: The CLI release builds no binaries: seven of eight legs fail

lexoliu:
Three changes, one of which the next release depends on.

* **#541** repairs `build-cli.yml`. The 0.2.0 release built one binary of
  eight and skipped `Upload Release Assets`: the Linux legs installed no
  native libraries, aarch64 Linux cross-compiled with a `pkg-config` that
  cannot cross, the Windows legs had no `dxc`, and `refresh-apt-index`
  named a source file that does not exist on 22.04, which resolved an
  empty index and took both WPE legs with it. The `kit` submodule also
  moves onto water-rs/waterkit#52, which is what makes the codec kit
  compile for x86_64 macOS. Fixes #540.
* **#542** gives `Photo` a content mode and has `Avatar` ask for
  `ContentMode::Fill`, so a non-square picture is cover-cropped instead
  of stretched. Fixes #427.
* **#543** corrects what `ci.yml` and `test.yml` say about themselves:
  since #458 the pull-request gate runs format and clippy only, and both
  files still described the older arrangement in which it ran tests.

Merging this is what lets the next release actually produce `water`
binaries: the CLI build workflow only runs behind a `waterui-cli-v*` tag,
so the fix has to be on `main` before the release run tags one.

### DIFF
--- a/.github/actions/refresh-apt-index/action.yml
+++ b/.github/actions/refresh-apt-index/action.yml
@@ -13,9 +13,21 @@ runs:
   steps:
     - name: Refresh the Ubuntu package index
       shell: bash
-      # Ubuntu 24.04 declares its archive in deb822 form at this path; apt
-      # picks the parser from the `.sources` extension.
-      run: >
-        sudo apt-get update
-        -o Dir::Etc::SourceList=/etc/apt/sources.list.d/ubuntu.sources
-        -o Dir::Etc::SourceParts=/dev/null
+      # Ubuntu 24.04 declares its archive in deb822 form at
+      # `sources.list.d/ubuntu.sources`; 22.04 still uses the one-line
+      # `sources.list`, and the runners here are both. Naming the file that is
+      # not there resolves an empty index instead of failing, so apt then
+      # reports every package as unlocatable and the install step aborts —
+      # which is how the WPE runtime legs of the CLI release died.
+      run: |
+        if [ -f /etc/apt/sources.list.d/ubuntu.sources ]; then
+          list=/etc/apt/sources.list.d/ubuntu.sources
+        elif [ -s /etc/apt/sources.list ]; then
+          list=/etc/apt/sources.list
+        else
+          echo "::error::no Ubuntu archive source file on this runner"
+          exit 1
+        fi
+        sudo apt-get update \
+          -o Dir::Etc::SourceList="$list" \
+          -o Dir::Etc::SourceParts=/dev/null
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -59,7 +59,7 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
           - target: aarch64-unknown-linux-gnu
-            os: ubuntu-latest
+            os: ubuntu-24.04-arm
           - target: x86_64-apple-darwin
             os: macos-latest
           - target: aarch64-apple-darwin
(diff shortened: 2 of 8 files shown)


lexoliu:
`Local API vs published baseline` is red on the baseline side, not on this diff: it fails building rustdoc for the published `waterui` 0.4.0, whose graph pulls a registry `waterui-chart` that still requires graphics 0.3 beside the 0.4 one. That is #538, pre-existing and immutable — the published baseline cannot be repaired retroactively. It clears once the next version publishes with the updated requirements, which is the release PR waiting behind this merge. The bench legs are informational.

PR #551, opened 2026-09-11
build(backends): consume hydrolysis 0.2.0 from crates.io and drop the in-tree copy

### LINKED ISSUE #480: Extract Hydrolysis into an independently versioned backend repository
Hydrolysis still inherits the WaterUI workspace and consumes repository-relative fixtures and dependencies. Extract backends/hydrolysis with traceable source history into its own repository, owning renderer implementation, dedicated examples, assets, native and GPU tests, CI, and releases. Replace workspace and CLI path assumptions with public package contracts and explicit compatible channel metadata. Preserve the GPU-required rendering model. Verify the standalone backend and consuming WaterUI integration before removing the in-tree implementation. The Material 3 theme must not be included in this repository: it has its own independent repository boundary.

lexoliu:
Fixes #480

Hydrolysis has its own repository (water-rs/hydrolysis) and its own release line; 0.2.0 is on crates.io, published from that repository against the released framework crates. This removes the in-tree renderer and consumes the published one. Stacked on #550, which did the same for the Material 3 theme.

## The patch table is what makes it one graph

A published `hydrolysis` depends on the registry `waterui`, `waterui-assets`, `waterui-backend-core`, `waterui-controls`, `waterui-core`, `waterui-form`, `waterui-graphics`, `waterui-icon`, `waterui-inspector-protocol`, `waterui-layout`, `waterui-locale`, `waterui-map`, `waterui-shape`, `waterui-text`, `waterui-webview` and `waterkit-clipboard`. Each one needs a `[patch.crates-io]` entry pointing at this tree, or the workspace carries two copies of it and a `View` handed to the renderer is not the `View` the renderer accepts. The theme change added several; the remainder arrive here.

The lock resolves with exactly one package added and no duplicate framework crate at any version.

## CI

Three legs named `-p hydrolysis` and no longer can:

- the all-features clippy and test sweeps excluded it, because `--all-features` asks for `webview-system` on every target; the exclusion and the per-feature steps that stood in for it both go, and the hydrolysis repository runs that sweep on Linux, macOS and Windows;
- the wasm32 leg linted `hydrolysis --features web` for the browser target. That repository has a wasm32 clippy leg of its own, so this one keeps the two crates whose wasm-only paths nothing else compiles (`waterui-media`, `waterkit-dialog`);
- the macOS `real_engine_macos` suite drove a real WKWebView through `webview-system`. That feature had no coverage at all in its new home — which is how a `compile_error!` that made the crate impossible to document reached a release run — so water-rs/hydrolysis#5 adds the leg there before it is dropped here.

`bench.yml` no longer watches two paths that do not exist; the benches still install the Material 3 theme and still render through Hydrolysis, both now as version requirements the manifest paths already cover.

## Verification

`cargo clippy --profile ci-check --workspace --exclude '*-example' --all-targets -- -D warnings` is clean. `cargo check -p waterui-testing --all-targets` builds the accessibility-tree host that every component's e2e test drives, which is the consumer that fails first if a patch entry is missing.


lexoliu:
Superseded by #553. Merging #550 with `--delete-branch` closed this one along with its base branch, and a closed pull request whose base is gone cannot be reopened or retargeted. #553 is the same change, rebuilt on `dev`.

Issue #481, opened 2026-09-09
Extract the Hydrolysis Material 3 theme into its own repository


lexoliu:
The hydrolysis-m3 package currently inherits the WaterUI workspace. Move backends/hydrolysis_m3 to a separate repository from both WaterUI and Hydrolysis, retaining traceable source history, theme implementation, assets, dedicated examples, visual and interaction tests, CI, and release ownership. Keep the theme on public widget and theme contracts, not renderer internals. Integrate through explicit compatible package versions and CLI channel metadata; do not add a submodule or bundle the theme implementation into the renderer repository. Verify standalone validation and WaterUI/Hydrolysis consumption before removing the in-tree copy.

PR #555, opened 2026-09-11
Merge dev into main: the Hydrolysis extraction, the dew seam, and the scaffold versions

### LINKED ISSUE #480: Extract Hydrolysis into an independently versioned backend repository

### LINKED ISSUE #548: Scaffolded projects pin framework versions one or two releases behind
`water create` scaffolds every project against framework versions that are one or two releases old. `cli/Cargo.toml`'s `[package.metadata.waterui-scaffold]` table is hand-written, nothing updates it when release-plz bumps a crate, and the only thing that would notice is a test that no longer runs on pull requests.

Measured on `dev` today, metadata against the in-tree version each entry claims to mirror:

| entry | scaffolded | actual |
| --- | --- | --- |
| `waterui-version` | 0.3.0 | 0.4.0 |
| `waterui-core-version` | 0.3.0 | 0.3.1 |
| `waterui-testing-version` | 0.3.0 | 0.4.0 |
| `waterui-ffi-version` | 0.3.0 | 0.3.1 |
| `waterui-dew-version` | 0.1.0 | 0.2.0 |
| `waterui-gtk-version` | 0.1.0 | 0.1.1 |
| `waterui-browser-cef-version` | 0.1.0 | 0.2.0 |
| `waterui-preview-version` | 0.1.0 | 0.1.1 |

Only `waterui-preview-protocol-version` still agrees, and only because that crate has not been released since.

`cli/build.rs` reads the workspace manifests when it can find them, so a CLI built from this checkout (`cargo install --path cli`) scaffolds against the real versions and the table is never consulted. The table is the fallback for a CLI that has no workspace around it — which is every CLI installed from crates.io. So the drift is invisible to us and lands on exactly the people who install the published tool: `cargo install waterui-cli` then `water create` produces an app pinned to `waterui = "0.3.0"` against a 0.4.0 framework, with a preview runtime a minor behind the protocol the CLI speaks.

That is also why the fallback has to be correct at publish time rather than merely eventually: the values baked into the released binary are whatever the committed table said when release-plz published it.

The guard exists — `fallback_release_versions_match_workspace_versions` in `cli/src/runtime/build_info.rs` asserts exactly this table against the workspace manifests. It is a plain `#[test]`, and since #458 moved every test leg to the nightly matrix nothing runs it on a pull request, so each release PR lands the drift unopposed and the nightly that would have caught it is red for other reasons.

Two things are needed: the values corrected, and the drift made impossible to land again — the version bump and the metadata that mirrors it have to travel in the same change, or the check that compares them has to sit in the per-change gate where a release PR meets it.

Found while extracting the Hydrolysis backend (#480); that change had to rewire the two hydrolysis entries and the drift in the rest of the table was visible beside them.

lexoliu:
Five changes, all green on `dev`:

- **#553 / #550** close #480 and #481. `hydrolysis` 0.2.0 and `hydrolysis-m3` 0.2.0 are published from their own repositories and consumed from crates.io; `backends/hydrolysis/` and `backends/hydrolysis_m3/` are gone. Sixteen `[patch.crates-io]` entries resolve the framework crates the published backends name to this tree, so the graph carries one copy of each type. The CI legs that named `-p hydrolysis` move to that repository, including the macOS WKWebView suite that was the only thing ever compiling `webview-system` (water-rs/hydrolysis#5, #6).
- **#554** closes #548. `[package.metadata.waterui-scaffold]` had drifted on every entry, so a CLI installed from crates.io scaffolded projects against `waterui = "0.3.0"` on a 0.4.0 framework. `cli/build.rs` now fails the build with the exact lines to write when a version falls behind — which means **a release PR is red until that table travels with the version bump**, deliberately.
- **#552** fixes dew's two bit-exact seam guards: chrome-carved children are placed from their `f64` region instead of through an `f32` frame, so a bar and the page beside it name the same edge whatever the font measures.
- **#549** clears the `cast_precision_loss` the avatar visual acceptance landed with.

Merging this runs CI on `main` and therefore the release workflow, which will regenerate the release PR without the two hydrolysis crates.


lexoliu:
`Local API vs published baseline` is red for the reason #538 records, and nothing in this branch touches it: cargo-semver-checks fails while building the **baseline**, which is the published `waterui` 0.4.0.

```
Building waterui v0.4.0 (current)   ->  Built [198.385s]
Building waterui v0.4.0 (baseline)  ->  error[E0277]: the trait bound
    `waterui_graphics::Color: Into<waterui_graphics::color::Color>` is not satisfied
    --> registry/.../waterui-chart-0.1.0/src/legend.rs:141:56
    --> registry/.../waterui-graphics-0.3.0/src/color/mod.rs:221:1
```

0.4.0 requires the extracted crates at their pre-foundation-0.4 versions, which require `waterui-graphics ^0.3`, so that baseline's own graph carries graphics 0.3 beside 0.4 and cannot compile. It is a property of what is already published and clears when 0.4.1 becomes the baseline — which is what merging this enables. Every other check is green.

Issue #519, opened 2026-09-10
apple: present GpuSurface and AppliedFilter output through IOSurface-backed textures on CALayer.contents instead of one CAMetalLayer swapchain per view


lexoliu:
On Apple every `GpuSurface` and every `AppliedFilter` output owns a `CAMetalLayer`, so each one carries its own wgpu surface and drawable pool (Fifo, `desired_maximum_frame_latency` 2) and presents on its own. Measured in #172: ~2.4 MB idle per extra surface on macOS, swapchain memory scaling with surface area (a full-window HDR surface is ~98 MB of drawables), and one submit + present per animated surface per frame (5.1 % CPU at one surface, 50.8 % at 56).

The presentation boundary can be a texture instead of a swapchain: render into an `IOSurface`-backed wgpu texture (wgpu creates Metal textures from an `IOSurface` through `MTLDevice.makeTexture(descriptor:iosurface:plane:)`, imported with `create_texture_from_hal::<Metal>`) and hand that `IOSurface` to `CALayer.contents`. Core Animation composites it in place, zero-copy, with no per-surface drawable pool and no `present`; a surface that did not change costs nothing. Double-buffer two `IOSurface`s per view and swap `contents` after the frame's completion handler so a frame in flight is never displayed.

Scope: `waterui_gpu_surface_*` and `waterui_applied_filter_*` on Apple (both `attach` paths, `WuiGpuSurfaceView`, `WuiAppliedFilter`), HDR through `IOSurface` pixel formats that match the current surface-format preference, and the same before/after measurement as #172 on `examples/icons` variants. The shared per-window rasterization boundary in #172 is a separate step; this one removes the swapchains without changing who rasterizes what.

lexoliu:
Reopening: water-rs/waterui#571 carried only the `waterui_applied_filter_*` half of this issue's scope and closed it on merge. `waterui_gpu_surface_*` — `WuiGpuSurfaceView`'s own `CAMetalLayer`, which is the larger share of the swapchain memory measured in water-rs/waterui#172 — is still on the swapchain, and the before/after measurement on `examples/icons` this issue asks for has not been taken. Both land in the PR that closes this.

lexoliu:
### Idle footprint, and why it does not settle this

[...]

PR #580, opened 2026-09-12
fix(ffi): give ViewEffect a host-texture path on Apple

### LINKED ISSUE #579: apple: a ViewEffect's output is invisible to every capture path, because it still presents through a CAMetalLayer
`ViewEffect` is the escape hatch for an application's own wgpu effect over a view, and on Apple its output reaches nobody. Nothing in this repository exercised it, which is why #519 converted `GpuSurface` and `AppliedFilter` to an `IOSurface`-backed presenter and left this third host on the `CAMetalLayer` it shares the defect with.

`WuiViewEffect` creates a `CAMetalLayer` and adds it as a sublayer:

```swift
private var outputLayer: CAMetalLayer!
...
let outputLayer = CAMetalLayer()
outputLayer.device = metalDevice
```

and `waterui_view_effect_attach` configures a wgpu `Surface` on it. A `CAMetalLayer`'s drawable is readable only by the pipeline that presented it, so it is invisible to `CALayer.render(in:)`, `cacheDisplay(in:to:)` and `CARenderer` — every path `water preview`, `WuiViewRenderer` and an enclosing `AppliedFilter` capture with. It is also forced under the native content it should sit above.

### Reproduction

A `#[preview]` in `examples/filter` wrapping the existing swatch grid in a `ViewEffect` whose renderer ignores its input and clears the output to opaque magenta — a colour nothing else in the view draws, so the image cannot be ambiguous about where the pixels went:

```rust
ViewEffect::new(sample_content(), FillEffect).size(220.0, 140.0)
```

`water preview view_effect_preview --platform macos` renders the heading and **nothing else**: not the fill, and not the content underneath it either, since the invisible output layer covers that too.

### Fix

The same conversion #519 made for the other two hosts. On Apple the host owns a pair of `IOSurface`-backed textures and hands one in per frame (`WuiSurfacePresenter`), the FFI grows a host-texture attach and render path beside the swapchain one that every other platform keeps, and the `CAMetalLayer` goes away. The preview above stays as the regression test.

lexoliu:
A `ViewEffect` rendered into a wgpu swapchain and the Apple backend presented it through a `CAMetalLayer`. A Metal drawable is readable only by the pipeline that presented it, so the effect's output was invisible to every capture path — the preview snapshot, `WuiViewRenderer`, and an enclosing filter all saw nothing where it drew, and the presentation was composited under the content it draws over.

This is the split #519 made for the other two GPU hosts, now applied to the third.

The attached output becomes a `ViewEffectTarget`: a surface and its configuration everywhere a swapchain is the right answer, and on Apple the output format alone, because there the host owns the textures. Beside the swapchain `attach`/`render` pair, Apple gets `waterui_view_effect_attach_host_textures`, `waterui_view_effect_resolve_output_size`, `waterui_view_effect_output_metal_pixel_format`, and `waterui_view_effect_render_to_metal_texture`. Each platform's other pair is a panicking twin rather than a no-op, so calling the wrong one is a crash and not a silently empty frame.

The Apple half is water-rs/apple-backend#26; the regenerated header also lands in water-rs/android-backend#16, whose bindings are otherwise unchanged.

### Regression test

`examples/filter` gains `view_effect_preview`: a `ViewEffect` that fills its output with opaque magenta over the swatch grid, which is deliberately unambiguous — the grid underneath is entirely covered, so the image answers one question, whether the effect's output reaches the screen and the capture at all. Before this change it rendered the heading and nothing else: not the fill, and not the content underneath it either.

`nested_filter_preview` and `filter_preview` render unchanged.

Fixes #579

### DIFF
--- a/examples/filter/src/lib.rs
+++ b/examples/filter/src/lib.rs
@@ -15,6 +15,9 @@
 use core::time::Duration;
 use waterui::animation::Animation;
 use waterui::app::App;
+use waterui::graphics::{
+    EffectRenderer, ViewEffect, ViewEffectContext, ViewEffectInput, ViewEffectOutput, wgpu,
+};
 use waterui::prelude::slider::slider;
 use waterui::prelude::*;
 use waterui::preview;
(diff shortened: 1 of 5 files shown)


lexoliu:
Added `effect_in_filter_preview`: the same `FillEffect` behind a component boundary, inside a completely desaturating filter. It renders a uniform grey rectangle — the outer host captured the effect's output, not the swatch grid underneath it and not an unfiltered magenta. That is the other half of what the `CAMetalLayer` drawable hid.

Issue #612, opened 2026-09-13
Extract waterui-gtk into a standalone repository with dedicated CI


lexoliu:
## Intent

The GTK backend used to live in-tree at `backends/gtk` as a workspace member. It now has its own repository — `water-rs/gtk-backend` — and this repo consumes it as a published crate, per the repository-boundary policy on `dev` (extracted crates are consumed via package versions, not submodules; the earlier submodule sketch below is superseded).

## Motivation

GTK was the only in-tree platform-native backend, and its coverage here was thin: the only tests that exercised it were `shape_geometry`'s pure-arithmetic unit tests — nothing ever initialized GTK4, snapshotted a widget, or touched the GL surface paths in CI. A dedicated repository gets CI that actually runs the backend: `gtk4` system dependencies, `xvfb` + llvmpipe for a real GL context on a headless runner, `cargo nextest` with GTK initialized.

## Scope

- New repo `water-rs/gtk-backend`, seeded from `backends/gtk` with history preserved (`git subtree split`), default branch `dev`. ✅
- Standalone `Cargo.toml`: no workspace inheritance — explicit dep versions against the published `waterui-*` crates, inline lints copied from the root workspace lints. ✅
- CI in the new repo: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, tests under `xvfb` with mesa/llvmpipe (default and `--all-features`), MSRV, non-Linux stub checks. ✅
- Publishing moves to the new repo (`release-plz`; the `waterui-gtk` crate name is already claimed on crates.io at 0.1.2, so no manual first publish needed, only a trusted-publisher entry).
- In this repo: `backends/gtk` is removed from workspace members and the tree; `waterui-gtk = "0.1.2"` becomes a `[workspace.dependencies]` requirement; the CLI scaffold keeps the framework-channel `=<version>` pin via `EXTERNAL_PACKAGES` + `waterui-gtk-version` metadata (mirroring the dew extraction, #642).

## Notes

`waterui-gtk` has no in-workspace consumers — the CLI emits it into user projects. The split therefore changes CI coverage location, not any API surface. Dev-channel scaffolding for lock-absent extracted crates is covered by the `resolve_packages` fallback in the consuming PR.

PR #643, opened 2026-09-13
build(gtk): consume the extracted gtk-backend repository as a submodule

lexoliu:
## Summary

Consumes `backends/gtk` as the extracted `water-rs/gtk-backend` submodule (tracking `dev`), completing the repo split from #612:

- `backends/gtk` in-tree sources → gitlink to `water-rs/gtk-backend` (same pattern as `backends/android` / `backends/apple`)
- Dropped from workspace `members` and `[workspace.dependencies]`: unlike android/apple the crate is published, but its standalone manifest pins released `waterui-*` versions (waterui 0.4.x) that this checkout's 0.3.0 manifests cannot satisfy — a member build would link a second registry copy of `waterui` beside the local one
- CLI gtk4 scaffolding emits the published version in every mode (`waterui-gtk-version` → `0.1.2`); the `backends/gtk` path kind is gone since a path dep on the submodule would hit the same duplicate-`waterui` problem
- `build_info` test no longer equates the scaffold pin with the in-tree manifest — the submodule's dev version intentionally leads the published pin

Also fixes a pre-existing test bug surfaced here: `gtk4_scaffold_uses_embedded_workspace_version` asserted `version = "x.y.z"` while the generator emits `version = "^x.y.z"` — it could never pass.

Note: `Cargo.lock` drops the `waterui-gtk` member entry and the gtk4/glib stack; the lockfile also self-healed a stale `waterui-canvas` resolution that was inconsistent with the git pin in `Cargo.toml`.

#### Test plan
- [x] `cargo check -p waterui-cli` clean
- [x] `cargo test -p waterui-cli gtk4` — 6+3 pass including the fixed scaffold test
- [x] `cargo metadata` regenerates a consistent lockfile
- [ ] workspace CI green

Refs #612


lexoliu:
Superseded by #646 — the repository-boundary policy on dev now extracts crates to published package versions rather than submodules.

Issue #711, opened 2026-09-13
Launch screen for every WaterUI backend


lexoliu:
## Intent

Every WaterUI backend should be able to show a launch screen: what the user sees between tapping the icon and the app's first frame. The motivating case is Hydrolysis on the web, where the page stays blank white for the whole wasm download, WebGPU device creation, Vello pipeline compilation and font fetch — but the capability is declared once and projected onto every platform.

## Contract

A launch screen is displayed before any Rust code runs (before the wasm is downloaded, before the dylib is loaded), so it is a build-time static asset, never a `View`. It is declared in `Water.toml` next to `[theme]` and `Icon.*`:

```toml
[launch]
background = "#0B1E3F"        # optional; defaults to theme.background, then the platform default
background_dark = "#000000"   # optional; defaults to background
```

The artwork is not a key: like the app icon (`Icon.*`), it is the single optional square `Launch.svg` / `Launch.png` at the root of the assets directory, discovered by convention.

The `water` CLI projects it onto each platform's native mechanism. The only runtime involvement is the "first frame presented" signal that ends it.

Defaults follow each platform's convention rather than one shared look:

| Platform | Without `[launch]` | With `Launch.*` |
| --- | --- | --- |
| iOS | system background color, no image (HIG: a launch screen resembles the first screen) | centered, respects safe area |
| Android | OS default: launcher icon on the window background | goes through the adaptive-icon safe-zone renderer (Android masks the splash icon in a circle) |
| Web | theme background + app icon + download progress bar | replaces the icon |
| macOS / GTK / desktop Hydrolysis | nothing: the window is created with its content already laid out, so no surface exists before the first frame; the window background belongs to `[theme]` | ignored, documented |

## Decisions

- Desktop platforms take nothing from `[launch]`: macOS, GTK and desktop Hydrolysis all present the window with its first frame, and the window background is `[theme]`'s job, not a launch-screen setting. #716 is closed with documentation.
- Web without `[launch]` shows theme background, app icon and a determinate progress bar driven by the wasm fetch byte count.
- Android API 26–30 uses `androidx.core:core-splashscreen`; API 31+ uses the platform SplashScreen API through the same library. APK size is measured before and after.
- The web "first frame presented" signal is a `CustomEvent` dispatched by the Hydrolysis web runner on the canvas; Hydrolysis ships it first, the CLI adopts it with a version bump.

## Structure

[...]

Issue #699, opened 2026-09-13
cli: path-pinned projects reference the removed backends/hydrolysis{,_m3} checkout directories


lexoliu:
A project pinned to a framework checkout (`waterui_path` in `Water.toml`) gets its generated Hydrolysis backend manifest written with `hydrolysis` and `hydrolysis-m3` as path dependencies on `<checkout>/backends/hydrolysis` and `<checkout>/backends/hydrolysis_m3` (`cli/src/project_model/templates.rs`, `NativeBackendDependencyPathKind::BackendsSubdir("hydrolysis")` / `("hydrolysis_m3")`, used by both the native and the wasm dependency tables). Those directories no longer exist: both crates were extracted to water-rs/hydrolysis and water-rs/hydrolysis-m3 (#480, #481) and the workspace consumes them through `[workspace.dependencies]` plus the `[patch.crates-io]` git pins in the root `Cargo.toml`.

Consequence: every local-development workflow that builds the managed Hydrolysis backend against a checkout — `water run --backend hydrolysis`, `water preview --backend hydrolysis`, `water mcp` — fails to resolve the dependency in a path-pinned project. The monorepo examples and playgrounds are all path-pinned per AGENTS.md, so this is the default developer path. Reproduced while smoke-testing `water mcp` on a fresh playground pinned to a `dev` checkout; it only built after symlinking detached hydrolysis worktrees at the pinned revisions into `backends/`.

Fix: for a path-pinned project, resolve `hydrolysis` / `hydrolysis-m3` exactly as the checkout does — read the root `Cargo.toml`'s `[workspace.dependencies]` entry and any `[patch.crates-io]` override for the crate, and emit the same source (registry version, or `git` + `rev`) into the generated manifest. `BackendsSubdir` then has no remaining users and goes away; do not keep it as a fallback. The generated manifest for a non-pinned project (registry versions from `build_info`) is unchanged.

Verification: a unit test in `templates.rs` renders the Hydrolysis manifest for a path-pinned context against a root manifest fixture carrying a `[patch.crates-io] hydrolysis = { git, rev }` entry and asserts the emitted dependency is that git source; `water create /tmp/x --mode playground` with `waterui_path` pointing at the checkout followed by `water preview --backend hydrolysis <fn>` builds without any `backends/hydrolysis*` directory present.

Issue #62, opened 2025-12-11
Size optimization


lexoliu:
![Image](https://github.com/user-attachments/assets/ed4b487a-050b-4740-8469-9a149843be26)
The size of WaterUII had been ridiculously large. We may require to take actions to reduce its size under 30MB..

lexoliu:
Also, we would do not provide a fat APK by default anymore. Users have to use `--arch` flag for `water package` to choose an architecture or architectures they want. It significantly reduces size. Now whole APK is only 39MB

lexoliu:
We removed deprecated Compose support, now whole app is ONLY 11MB. Excellent!

PR #740, opened 2026-09-13
feat(cli): launch screen for the Hydrolysis web page

### LINKED ISSUE #715: Launch screen and startup measurement for the Hydrolysis web bundle
Give the Hydrolysis web bundle a launch screen, and measure and fix the startup path it covers.

## Current behaviour

`index.html` renders a blank white page (`style.css` hard-codes `background: #ffffff`) from navigation until the first Vello frame. In between: wasm download (the CLI dev server sends no `Content-Encoding`), `WebAssembly.instantiate`, WebGPU adapter/device request, Vello pipeline compilation, and only then the font manifest and every font file fetched one after another (`hydrolysis/src/runner/web_runner.rs`, `load_web_fonts` after `HydrolysisRenderer::new`).

## Scope

1. **Measure first.** Instrument the startup timeline with `performance.mark` in `bootstrap.js` and `tracing` on the Rust side: wasm bytes, fetch, instantiate, adapter, device, pipeline compilation, fonts, first frame. Record the numbers in the pull request; the launch screen does not make the page faster and the numbers say what does.
2. **Launch screen in `index.html`**: an inline `#waterui-launch` layer painted before the wasm fetch starts — background per `prefers-color-scheme` from the resolved `[launch]` values, the launch image (or the app icon by default) inlined as SVG or a data URI so it costs no request, and a progress bar.
3. **Progress**: `bootstrap.js` fetches the wasm itself, counts bytes through a `ReadableStream` (determinate when `Content-Length` is present — the CLI dev server always sends it), and hands a `Response` with the original headers to wasm-bindgen's `init()` so `instantiateStreaming` still applies.
4. **End signal**: the Hydrolysis web runner dispatches `CustomEvent("waterui:first-frame")` on the canvas after the first `Refresh` frame is presented; `bootstrap.js` removes the launch layer on it. This is a water-rs/hydrolysis pull request that ships first (the event is harmless without a listener); this repository then bumps `hydrolysis-version` and adopts it.
5. **Startup fixes the measurement is expected to confirm** (Hydrolysis side): fetch fonts concurrently with adapter/device/pipeline setup and concurrently with each other (water-rs/hydrolysis#21). Dev-server compression is out of scope here: the dev server exists for local iteration where the wasm comes off loopback, and a production deployment's server owns its `Content-Encoding`; the launch page stays determinate either way because the CLI stamps the bundle's byte size into `index.html` rather than relying on `Content-Length`.

## Blockers for the end-to-end measurement

A Hydrolysis web bundle does not currently build at all, so the timeline in scope item 1 and the screenshots below wait on:

- water-rs/waterkit#62 — `waterkit-audio` failed to compile for `wasm32-unknown-unknown`; fixed on waterkit `dev` by water-rs/waterkit#63 and reaches this tree with the next waterkit release.
- #699 — path-pinned projects still reference the removed `backends/hydrolysis` checkout.

## Verification

Screenshots at desktop and phone widths of the launch layer mid-download and of the handoff to the first frame; the measured timeline before and after.

lexoliu:
Fixes #715. Part of #711. Stacked on #717 (`feat/launch-config`); retarget to `dev` once that merges.

## What

- `index.html` is the launch screen: background per appearance from the resolved `[launch]` values (or the browser's own `Canvas`/`CanvasText` pair, which follows the system appearance by itself), the `Launch.*` artwork or the app icon inlined as a PNG data URI, and a progress bar — all inline, painted before the first fetch. `style.css` no longer hard-codes a white page.
- `bootstrap.js` fetches the wasm itself and counts bytes through a `ReadableStream` for a determinate bar; the CLI stamps the bundle's byte size into the page (`data-wasm-bytes`) so the bar stays correct however the file is transferred (a compressed transfer's `Content-Length` is the compressed size while the stream yields decoded bytes). The counted stream is handed to wasm-bindgen's `init()` as a `Response` with the original headers, so `instantiateStreaming` still applies.
- The layer stands down on the `waterui:first-frame` event the Hydrolysis web runner dispatches on the canvas — water-rs/hydrolysis#21, which also fetches fonts concurrently with adapter/device setup and with each other instead of one after another after pipeline compilation. That PR ships first (the event is inert without a listener); this PR's `hydrolysis-version` bump follows its release.
- The web shell templates move out of the scaffolded backend tree (`templates/hydrolysis_web/`): the backend project never contained a page, the page needs the built bundle, and it is the one template that interpolates user text into HTML, so it is HTML-escaped (`askama.toml` escaper for `.tpl`; every scaffold template already declares `escape = "none"` explicitly).
- `LaunchAssets` in staging: the resolved plan plus the loaded artwork, shared by the per-platform sub-issues. `HexColor::relative_luminance` picks the progress bar's foreground against a configured background.

## Not verified end to end yet

A Hydrolysis web bundle cannot be built today: water-rs/waterkit#62 (`waterkit-audio` does not compile for wasm32 and every Hydrolysis app links it through `video-gpu`) and #699 (path-pinned projects point at the removed `backends/hydrolysis`). Both block this issue on GitHub. The page rendering is covered by unit tests; the browser-side verification and the startup timeline measurement from #715 follow once a bundle builds.

## Checks

`cargo clippy -p waterui-cli --all-targets -D warnings`, `cargo clippy -p waterui-assets-planner --all-targets -D warnings`, `cargo nextest run -p waterui-cli -E 'test(/web_launch|assets|manifest|theme|colors_xml|icon|scaffold/)'` (51 passed).


lexoliu:
## Visual check of the shell

A Hydrolysis web bundle does not build yet (water-rs/waterkit#62, #699), so the shell was rendered through `write_web_shell` against `examples/shape` with a 3 MB placeholder `pkg/app_bg.wasm` and served at 500 KB/s to make the download visible. Looked at in the browser at desktop and 375 px widths, light and dark:

| Case | Seen |
| --- | --- |
| `[launch] background = "#0B1E3F"`, `background_dark = "#000000"`, `assets/Launch.svg` | navy page, artwork centered at 128 px, determinate bar advancing with the byte count; black in dark appearance |
| no `[launch]`, no `Launch.*` | `Canvas` background follows the system appearance (white / `#121212`), app icon in place of the artwork, same bar |
| `waterui:first-frame` dispatched on the canvas | layer fades over 220 ms and is removed from the DOM |
| `init()` rejects | error text replaces the page on the launch background |

The last case showed a defect: with a configured background the error text kept the default text color (black on navy). `html, body` now carry `color: var(--waterui-launch-foreground)` so the message reads on whatever background the launch screen chose; folded into the commit.

PR #742, opened 2026-09-13
feat(cli): project the launch screen onto iOS through UILaunchScreen

### LINKED ISSUE #713: Project [launch] onto iOS UILaunchScreen and the macOS window background
Project `[launch]` onto iOS through `UILaunchScreen`, and set the macOS window background color.

## iOS

- Stage `LaunchBackground.colorset` (universal + dark appearance from `background` / `background_dark`) and, when `image` is set, `LaunchImage.imageset` at 1x/2x/3x into the app's xcassets alongside the existing `AppIcon` staging in `cli/src/project_model/assets/unified.rs`.
- The generated `project.pbxproj` already sets `INFOPLIST_KEY_UILaunchScreen_Generation = YES`; add `INFOPLIST_KEY_UILaunchScreen_UIColorName`, `INFOPLIST_KEY_UILaunchScreen_UIImageName` and `INFOPLIST_KEY_UILaunchScreen_UIImageRespectsSafeAreaInsets = YES` from the template context, only when the corresponding value is configured, so an unconfigured project keeps the system-generated launch screen.
- No Rust or Swift runtime involvement: the OS removes the launch screen at the app's first presented frame, and the WaterUI tree is built synchronously in `didFinishLaunchingWithOptions`.

## macOS

- `NSWindow.backgroundColor` in the generated `AppNameApp.swift` takes the resolved launch background so the window never flashes white before the first frame. macOS has no launch screen; this is documented in the skill, not emulated.

## Verification

Screenshots of the iOS simulator launch (with and without `[launch]`, light and dark) and of the macOS window's first frame.

lexoliu:
Fixes #713. Part of #711. Stacked on #740 (`feat/launch-web`); retarget once that merges.

## iOS

- Staging writes `LaunchBackground.colorset` (universal color plus a `luminosity: dark` appearance when the dark background differs) and, when `Launch.*` exists, `LaunchImage.imageset` at 1x/2x/3x of a 128 pt box, next to the existing `AppIcon`/`AccentColor` sets. `AccentColor` goes through the same color-set writer.
- The generated `project.pbxproj` sets `INFOPLIST_KEY_UILaunchScreen_UIColorName = LaunchBackground` and `UIImageName = LaunchImage` + `UIImageRespectsSafeAreaInsets = YES` only for what was staged (`LaunchTemplateEntry` on the template context, resolved in `AppleBackend::init` from the same `LaunchAssets` staging uses), so an unconfigured project keeps the system-generated launch screen, as before.
- No runtime involvement: iOS removes the launch screen at the first presented frame, and the WaterUI tree is built synchronously in `didFinishLaunchingWithOptions`.

## macOS

Takes nothing from `[launch]`. The window is created after the tree exists, on the system window background; there is no moment before the first frame for a launch color to show, and painting one into `NSWindow.backgroundColor` would leak launch configuration into the window's steady-state appearance, which belongs to `[theme]`. The skill documents desktop platforms this way (docs commit on #717).

## Checks

`cargo clippy -p waterui-cli --all-targets -D warnings`; `cargo nextest run -p waterui-cli -E 'test(/apple|launch|assets|colors|icon|theme|scaffold/)'` (71 passed), including the color-set/image-set writers and the pbxproj rendering with and without launch assets. Simulator screenshots (light/dark, with and without `[launch]`) are being collected separately and will be posted here.


lexoliu:
## Simulator check (iPhone 16 Pro, iOS 26.5), `examples/shape`

| Case | Seen at launch |
| --- | --- |
| no `[launch]`, no `Launch.*` | the system-generated launch screen: plain system background, no image |
| `background = "#0B1E3F"`, `background_dark = "#000000"`, `assets/Launch.svg`, light appearance | navy background, artwork centered inside the safe area |
| same project, dark appearance | black background, artwork centered |

The built `Info.plist` carries `UILaunchScreen → UIColorName = LaunchBackground, UIImageName = LaunchImage, UIImageRespectsSafeAreaInsets = true`; `Assets.car` holds `LaunchBackground` (any + dark appearance) and `LaunchImage` at 1x/2x/3x.

The check caught two defects in the first version of this branch, both now fixed in the commit:

1. `INFOPLIST_KEY_UILaunchScreen_UIColorName` / `UIImageName` / `UIImageRespectsSafeAreaInsets` are not Xcode build settings — Xcode's specs register only `INFOPLIST_KEY_UILaunchScreen_Generation` and `INFOPLIST_KEY_UILaunchStoryboardName`, and unregistered `INFOPLIST_KEY_*` values are dropped silently — so the app shipped an empty `UILaunchScreen` dictionary and a plain launch screen. The scaffold now writes `<App>/Info.plist` from `cli/src/templates/apple/AppName/Info.plist.tpl` with the `UILaunchScreen` dictionary (empty when nothing is configured, which is what `_Generation = YES` produced), and the pbxproj sets `INFOPLIST_FILE` beside `GENERATE_INFOPLIST_FILE = YES` so Xcode merges the generated keys into it.
2. `<App>/` is a file-system-synchronized group, so that `Info.plist` was also copied as a bundle resource and every iOS build failed with `Multiple commands produce '….app/Info.plist'`. The group now carries a `PBXFileSystemSynchronizedBuildFileExceptionSet` with `membershipExceptions = (Info.plist)`, which is what Xcode itself writes for a plist inside a synchronized folder.

Template tests cover both (`apple_project_names_only_the_launch_assets_that_were_staged`).

lexoliu:
This merge landed in the stack base only: `feat/launch-web` was force-pushed afterwards with a rebased web commit, which dropped this merge before #740 went into `dev`, so nothing from here reached `dev`. The same commit is re-landed on top of `dev` in #886.

PR #797, opened 2026-09-14
fix(cli): reject iOS simulators whose runtime is below the deployment target before building

### LINKED ISSUE #786: water run --platform ios accepts a simulator whose iOS runtime is below the deployment target and only fails at install
`find_device` in `cli/src/terminal/commands/run.rs` selects an iOS simulator by `-d <udid|name>` or, when none is given, the first booted (else first available) device from `AppleSimulator::scan_ios()`. It never compares the simulator's runtime with the app's `IPHONEOS_DEPLOYMENT_TARGET`, although the CLI already has both: `apple_deployment_target` reads the target from the scaffolded `project.pbxproj` (currently `26.0`), and every `AppleSimulator` carries `runtime_identifier` (`com.apple.CoreSimulator.SimRuntime.iOS-18-5`), with `xcrun simctl list -j` also listing each runtime's `version`.

Observed on `examples/shape` with an iOS 18.5 iPhone 16 Pro booted: `water run --platform ios -d <udid>` builds the Rust dylib and the Xcode project (over ten minutes cold) and then fails at `simctl install` with `IXUserPresentableErrorDomain code=16: "Shape Example" Requires a Newer Version of iOS — You need to update this iPhone to iOS 26.0`. The same happens without `-d` whenever the first booted simulator runs an older iOS.

Expected: the check happens before anything is built. A device named with `-d` whose runtime is below the deployment target is rejected with an error naming the simulator, its iOS version and the required target; automatic selection considers only simulators whose runtime satisfies the target and reports when none does. The runtime version comes from `simctl`'s runtime list (or the identifier), the target from the project as `apple_deployment_target` already derives it; neither is hardcoded. The same comparison should apply wherever a simulator is chosen for the app (`water run`, the preview and inspector launchers that call `scan_ios`).

lexoliu:
Fixes #786

## Cause

`find_device` (`cli/src/terminal/commands/run.rs`) selected an iOS simulator by `-d <udid|name>`, else the first booted, else the first available, and never compared its runtime with the app's `IPHONEOS_DEPLOYMENT_TARGET`. On this machine two simulators are both named "iPhone 16 Pro" (iOS 18.5 and 26.5); `-d` by name or automatic selection took the 18.5 one, the Rust dylib and the Xcode project built for over ten minutes, and `simctl install` then failed with `IXUserPresentableErrorDomain code=16: Requires a Newer Version of iOS`. Pre-existing; surfaced while verifying #742.

## Fix

- `AppleSimulator::scan` reads `xcrun simctl list --json` (the `devices`-only form carries no `runtimes` array) and attaches each runtime's `version` to its simulators as `runtime_version: Option<semver::Version>`; a runtime without a parseable version can never qualify.
- `AppleSimulator::select_ios(project, device)` reads the target through the existing `apple_deployment_target`, then: a `-d` match whose runtime is below the target is rejected naming the simulator, its UDID, its iOS version and the target; a name matching several qualifying simulators is an error listing the candidates by UDID; automatic selection takes the first booted qualifying simulator, else the first qualifying one, else an error naming the target and listing what exists.
- `water run`, `preview/launcher.rs` and `workflows/inspector/launcher.rs` all select through it; their three copies of "first booted else first" are gone.
- `parse_semver_version` (pads `26.0` to `26.0.0`) moves from `toolchain/rust.rs` to `runtime/utils.rs` so the toolchain check and the simulator gate share one parser.

## Verification

`cargo clippy -p waterui-cli --all-targets -- -D warnings` clean; `cargo nextest run -p waterui-cli -E 'test(apple) | test(simulator) | test(run)'` 72/72, including nine new tests on hand-built simulator lists (booted 18.5 + booted 26.5 + shutdown 26.5, below-target `-d`, ambiguous name, nothing qualifying) with no `simctl` shell-outs.


lexoliu:
The red `Test / lint` is a rustfmt diff in `components/visual/graphics/src/gradients/animated_mesh_gradient.rs`, pre-existing on `dev` since #796 and untouched here; fixed by #824. This pull request's own files are rustfmt-clean and clippy-clean.

PR #743, opened 2026-09-13
feat(cli): project the launch screen onto the Android SplashScreen API

lexoliu:
Fixes #714. Part of #711. Stacked on #742 (`feat/launch-apple`); retarget once that merges.

## What

- `androidx.core:core-splashscreen:1.2.0` in the generated app; `MainActivity.onCreate` calls `installSplashScreen()` before `super.onCreate`; the launcher activity's theme is `Theme.WaterUIApp.Launch` (`parent="Theme.SplashScreen"`, `postSplashScreenTheme = @style/Theme.WaterUIApp`). API 31+ uses the platform SplashScreen API through the same library; 26–30 get the backport.
- Resolved `[launch]` values: `windowSplashScreenBackground` → `@color/waterui_launch_background` (`values-night/colors.xml` overrides it when the dark background differs; the night file is removed otherwise); `Launch.*` → `windowSplashScreenAnimatedIcon` as an adaptive icon (`drawable/ic_launch_artwork.xml` over a safe-zone-placed foreground PNG and an edge-derived background color), because Android masks the splash icon to a circle. Without `[launch]` the theme carries only `postSplashScreenTheme`, so the OS default (launcher icon on the window background) stands.
- No keep-on-screen condition: the view tree is built synchronously in `onCreate`, so the platform's default end of the splash — the activity's first frame — is the app.
- `colors.xml` and `themes.xml` were built by string concatenation in `unified.rs` while `templates/android/.../values/` carried a second, unused copy of each. Both are now one askama template each (`templates/android_res/`) rendered by staging, and the scaffold copies are gone.

## Package size

Principle 5: release APK size before/after `core-splashscreen` is being measured and will be posted here.

## Checks

`cargo clippy -p waterui-cli --all-targets -D warnings`; `cargo nextest run -p waterui-cli -E 'test(/android|apple|launch|assets|colors|icon|theme|scaffold/)'` (102 passed), including the color table (launcher background derivation, launch colors, night override) and the theme rendering with and without launch background/artwork. Pixel 9 Pro screenshots follow.

PR #887, opened 2026-09-14
feat(cli): project the launch screen onto the Android SplashScreen API

### LINKED ISSUE #714: Project [launch] onto the Android SplashScreen API
Project `[launch]` onto Android through the SplashScreen API, with `androidx.core:core-splashscreen` covering API 26–30.

## Scope

- Add `androidx.core:core-splashscreen` to the generated app's `build.gradle.kts`; `MainActivity.onCreate` calls `installSplashScreen()` before `super.onCreate`. The activity theme becomes a `Theme.SplashScreen` child whose `postSplashScreenTheme` is `Theme.WaterUIApp`.
- Theme items from the resolved `[launch]` values: `windowSplashScreenBackground` (light and `values-night`), and `windowSplashScreenAnimatedIcon` when `image` is set — the image goes through the existing adaptive-icon safe-zone renderer in `assets/icon.rs`, because Android masks the splash icon in a circle. Without `image` the OS default (launcher icon) stands.
- The Android theme XML is currently built by string concatenation in `build_android_themes_xml` (`cli/src/project_model/assets/unified.rs`) while `templates/android/.../values/themes.xml.tpl` is a second, unused source of the same file. Replace both with one askama template rendered by staging, per the repository rule that structured text is never concatenated.
- No keep-on-screen condition is needed: the WaterUI view tree is built synchronously in `onCreate`, so the platform's default "dismiss on first frame" is the correct end of the launch screen.

## Package size

Principle 5: record the release APK size before and after adding `core-splashscreen` in the pull request.

## Verification

Screenshots on the Pixel 9 Pro (API 35) with and without `[launch]`, light and dark.

lexoliu:
Fixes #714. Part of #711. Stacked on the iOS pull request (`feat/launch-apple`); retarget to `dev` once it merges. Re-landing of #743, whose merge into the stack base never reached `dev`; identical content rebased.

## What

- `androidx.core:core-splashscreen:1.2.0` in the generated app; `MainActivity.onCreate` calls `installSplashScreen()` before `super.onCreate`; the launcher activity's theme is `Theme.WaterUIApp.Launch` (`parent="Theme.SplashScreen"`, `postSplashScreenTheme = @style/Theme.WaterUIApp`). API 31+ uses the platform SplashScreen API through the same library; 26–30 get the backport.
- Resolved `[launch]` values: `windowSplashScreenBackground` → `@color/waterui_launch_background` (`values-night/colors.xml` overrides it when the dark background differs; the night file is removed otherwise); `Launch.*` → `windowSplashScreenAnimatedIcon` as an adaptive icon (`drawable/ic_launch_artwork.xml` over a safe-zone-placed foreground PNG and an edge-derived background color), because Android masks the splash icon to a circle. Without `[launch]` the theme carries only `postSplashScreenTheme`, so the OS default (launcher icon on the window background) stands.
- No keep-on-screen condition: the view tree is built synchronously in `onCreate`, so the platform's default end of the splash — the activity's first frame — is the app.
- `colors.xml` and `themes.xml` were built by string concatenation in `unified.rs` while `templates/android/.../values/` carried a second, unused copy of each. Both are now one askama template each (`templates/android_res/`) rendered by staging, and the scaffold copies are gone.

## Package size

Principle 5: release APK size before/after `core-splashscreen` is being measured and will be posted here.

## Checks

`cargo clippy -p waterui-cli --all-targets -D warnings`; `cargo nextest run -p waterui-cli -E 'test(/android|apple|launch|assets|colors|icon|theme|scaffold/)'` (102 passed), including the color table (launcher background derivation, launch colors, night override) and the theme rendering with and without launch background/artwork. Pixel 9 Pro screenshots follow.


lexoliu:
Auto-closed when `feat/launch-apple` was deleted by the #886 merge; the same commit continues in #893 against `dev`.

PR #910, opened 2026-09-14
fix(cli): unify signal_dev_server_tree's mutability across cfg variants

lexoliu:
The unix variant takes `&Child` (killpg only needs the pid) while the `not(unix)` variant takes `&mut Child` (`kill` is `&mut self`), but all four call sites pass `&child` — so the CLI fails to compile on Windows. Both variants now take `&mut Child` and callers pass `&mut child`.


lexoliu:
Superseded: the CLI moved to water-rs/cli (#899), where this code was rewritten as DevServerTree (Windows job objects, uniform &self API) — the mutability mismatch no longer exists on either side.

PR #912, opened 2026-09-14
fix(apple): replay simulator log history after stream attach

### LINKED ISSUE #908: iOS simulator log stream misses launch-time markers emitted before attach
On iOS, `water run --logs info` never reports `waterui_first_paint_ms`: `AppleSimulator::run` calls `simctl launch` and only then spawns `log stream --predicate 'processID == pid AND subsystem == "dev.waterui"'` (`cli/src/apple/device.rs`, `start_log_stream`). The stream attaches to logd after launch, but `log stream` forwards only entries written after attach — the app can reach first paint in under a second on a warm simulator, so the marker is emitted before the stream exists and is permanently lost.

Evidence: apple-backend nightly run 34892897690 — every iOS example recorded `null` startup while launched examples still produced peak-RSS samples, and per-example iOS logs show zero `dev.waterui` lines after "Application started". The same race exists on the macOS path in principle (the stream starts after the process does); macOS just reaches first paint slowly enough that the attach usually wins — the occasional macOS null (e.g. `stress`) is consistent with the race.

A fast crash's `panic.payload` is lost the same way.

Fix: alongside the stream, run `log show --last <recent> --predicate <same> --style compact` once a few seconds after launch and forward those lines into the same event path — the persisted store retains pre-attach entries; consumers take the first matching marker so a duplicated stream line is harmless.

lexoliu:
## Summary

- `simctl launch` returns before `log stream` attaches to logd; the stream forwards only post-attach entries, so a fast first paint emits `waterui_first_paint_ms` into the void — every iOS example in nightly run 34892897690 recorded a null startup time while still producing memory samples, and macOS loses the occasional marker the same way
- A one-shot `log show --last 2m` replay of the persisted store forwards pre-attach entries through the same event path ~4s after the stream starts; consumers take the first matching marker, so a duplicated stream line is harmless
- Panic payloads emitted before attach are recovered through the same replay

#### Test plan
- [x] `cargo check`/`clippy -p waterui-cli` clean
- [ ] iOS e2e shards record non-null `waterui_first_paint_ms` on next aligned nightly

Fixes #908


lexoliu:
superseded — the CLI moved to water-rs/cli; fix landed there as water-rs/cli#6

PR #915, opened 2026-09-15
feat(cli): run on physical iOS devices via devicectl

lexoliu:
## Summary

`water run` gains physical-iPhone support: `devicectl` discovery (USB and "Connect via network"), `aarch64-apple-ios` builds, automatic signing with provisioning-profile minting, `devicectl device install` + `process launch --console` with environment delivery and stdio streaming, and remote termination on Ctrl+C.

- `cli/src/apple/physical.rs` — `ApplePhysicalDevice` model + list parser (fixture captured from a real iPhone 16 Pro), usability diagnosis (unreachable tunnel / not booted / Developer Mode off) with remedies, install→launch→stream→terminate lifecycle
- Selection: `--device` matches a paired device by name/identifier/UDID alongside simulators; without `--device`, a usable paired device is the fallback when no simulator qualifies
- Signing: device targets always sign — `CODE_SIGN_STYLE=Automatic`, `DEVELOPMENT_TEAM` resolved from Xcode's signed-in provisioning teams (keychain fallback), `-allowProvisioningUpdates`; `sign_staged_device_libraries` re-signs the Rust dylibs copied into `Frameworks/` after `xcodebuild` (dyld refused them unsigned — "missing code signature")
- Dev server: physical devices get the URL rewritten to the Mac's LAN IPv4 and the dev server binds non-loopback; `WATERUI_DEV_URL` goes through `devicectl -e` plus a `--waterui-dev-url=` launch-arg fallback
- `--logs`: on Apple targets `WATERUI_LOG` output is duplicated to stderr so it reaches `devicectl --console` (no host-side `log stream --device-udid` on this toolchain)
- `water devices` lists physical devices with transport, OS version, usability state
- Pins `apple-backend` to `0.3.0-dev.2` — `0.3.0-dev.1` sat on a stale lineage missing dev's FFI capability gates, which broke linking for any app not enabling the gated features

#### Test plan

- [x] `cargo clippy -p waterui-cli --all-targets -D warnings`, `cargo fmt`, `cargo test -p waterui-cli` (364+54)
- [x] `water devices` lists the paired iPhone (USB transport, iOS 27)
- [x] `water run --platform ios --device <id>` on examples/gallery: built, signed (fresh `iOS Team Provisioning Profile` covering the device), installed, launched — app ran ~27s with live stderr streaming, crash detected (`signal 6`), clean exit
- [x] Direct `devicectl launch` re-probe after on-device developer trust

Fixes #905.

PR #918, opened 2026-09-15
feat(cli): remember last-used run device per target

### LINKED ISSUE #911: run: remember last-used device per target; hydrolysis default on Linux
## Background

`water run` already resolves a default platform (host OS) and a per-platform default backend, and has a `--device` flag documented as "if not specified, uses first available device". Two gaps:

1. **Default backend table is stale.** Linux currently defaults to `gtk4`; the intended default is hydrolysis (the MD3 renderer), with gtk4 kept but **marked experimental**. Windows defaults to hydrolysis today (a future winui backend would take over as default when it lands).
2. **No device selection UX.** When a target has multiple devices (iOS simulators + physical devices from #905, Android devices/emulators), the CLI silently picks "first available" — an unstable, arbitrary ordering. The user has no say, and no memory of a previous say.

### Default backends (unchanged where already correct)

| Host | Default |
|---|---|
| macOS | apple + macOS app |
| Linux | hydrolysis (MD3); `gtk4` selectable but marked experimental |
| Windows | hydrolysis (winui when it exists) |

`--platform`/`--backend` continue to override (e.g. `--platform ios --backend apple`).

### Device selection

For targets with a device dimension (apple+iOS, android):

- `--device <id>` given → use it, **persist as last-used**.
- No `--device`:
  - Remembered device for this `(backend, platform)` still present in scan → use it.
  - Remembered device absent → interactive: warn + re-prompt; non-interactive: clear error listing current devices.
  - Exactly one device and nothing remembered → use it (no prompt, nothing persisted).
  - Multiple devices, nothing remembered → interactive picker (`dialoguer::Select`, matches existing `Confirm` usage) → persist the choice.
- Persisted state lives in `~/.water/config.toml` (`WaterConfig`), keyed by `(backend, platform)` so an iOS pick never leaks to Android. Value = the device's stable identifier (simulator UDID / device UDID).

### Local-only targets

`apple+macos`, gtk4, hydrolysis desktop, and web have exactly one implicit device: the host machine. `--device` must be **rejected** there — today `validate_device_arg` only rejects web and gtk4/hydrolysis; `apple+macos` silently ignores it.

## Acceptance

- `water run` on macOS still runs the macOS app with zero prompts.
- `water run --platform ios` with multiple simulators/devices prompts once, remembers, and reuses; `--device` overrides and becomes the new remembered device.
- `water run --platform macos --device foo` errors clearly.
- `water run --platform linux` uses hydrolysis; `--backend gtk4` works but warns "experimental".
- Non-interactive multi-device runs fail with a device list instead of prompting.
- Unit tests: config persistence round-trip, selection precedence (flag > remembered > picker), local-only rejection.

lexoliu:
## Summary
- `water run` on multi-device targets resolves the device as **--device > remembered > single candidate > interactive picker**, persisting the last-used stable identifier per `"<backend>/<platform>"` in `~/.water/config.toml` (`[last_used_device]`).
- iOS candidates span qualifying simulators (booted first) and usable paired devices; Android candidates span connected devices and AVDs (`--device` now also accepts an AVD name).
- Non-interactive runs with multiple candidates fail listing every device instead of hanging on a prompt; a stale remembered device warns and falls through to single/pick.
- Local-machine targets (apple/macos, gtk4, hydrolysis, web) reject `--device` outright instead of ignoring it.
- Linux defaults to hydrolysis; GTK4 stays selectable and is warned as experimental. Windows keeps hydrolysis until a WinUI backend exists.

#### Test plan
- [x] `cargo test -p waterui-cli` (config round-trip, `device_choice` precedence, local-only rejection, non-interactive listing)
- [x] `cargo clippy -p waterui-cli --all-targets -D warnings`, `cargo fmt --check`
- [x] Live: `water run --platform ios` non-interactively lists the booted sim + all shutdown sims; seeded config picks the remembered sim without prompting; `--device "iPhone 17 Pro"` persists its UDID; `water run --device foo` on macOS fails fast
- [ ] Interactive picker on a real TTY (paths verified via `device_choice`/`prompt_for_device` unit coverage)

Stacked on #915 (uses `ApplePhysicalDevice`); the diff shrinks to this change once that merges.

Fixes #911


lexoliu:
Retargeted: the CLI moved to water-rs/cli in cbe51d7d9. This change is now water-rs/cli#8.

PR #935, opened 2026-09-15
ci: restore the nightly full matrix and gate releases on its certification

### LINKED ISSUE #923: test.yml's full matrix has no caller — framework tests never run
`test.yml` is `workflow_call`-only, and every job except `lint` is gated on `inputs.full`. The only callers are `ci.yml` and `dev.yml`, both passing `full: false`. No trigger — PR, `dev` push, or schedule — ever runs the workspace `cargo nextest` pass, doctests, the wasm32 lint, the all-features passes, `macos-suites`, or the examples tests.

The comments attribute coverage to the per-repository backend nightlies (#755), but those drive example e2e launches, not the framework's own unit test suite. A PR that breaks a `waterui-core` unit test merges green; the skipped jobs still appear on PRs as `skipping` checks, which reads as coverage that does not exist.

Either the `full` matrix needs a scheduled caller (a nightly), or the dormant jobs are ~470 lines of CI carrying an illusion of coverage and should be removed.

Found while auditing the workflows; likely an intentional #458/#755 tradeoff worth recording either way.

lexoliu:
After #756 retired `nightly.yml`, the framework's own suite ran nowhere: the backend nightlies drive example e2e on their platforms and never execute this workspace's 875 test functions (180 files, plus 39 `tests/*.rs` targets), and `test.yml`'s `full` input had no caller. The fast gate's clippy was `--lib --bins`, so a `#[cfg(test)]` module that stopped compiling merged green.

This restores the nightly and narrows the gate to the release boundary:

- `nightly.yml` calls `test.yml` with `full: true` plus the Windows and web view JavaScript lanes, coverage and `cargo hack --each-feature`, daily at 08:11 UTC (after the backend nightlies, 02:00–06:47 UTC, have left the org's job cap) and on dispatch. A red run opens or updates one tracking issue.
- A green run certifies its commit as an immutable `nightly-*` prerelease. The manifest comes from a new `nightly` subcommand on `framework_manifest.py`, so it carries schema 2 like the stable one and what `water channel nightly` reads; `prepare_nightly.py` (schema 1, read the since-extracted `cli/Cargo.toml`) stays deleted.
- `Nightly healthy` in `ci.yml` requires that certification only on pull requests into `main`. A red nightly blocks a release, never integration on `dev`, so the deadlock #756 cited cannot occur and the `nightly-fix` label path is gone.
- The fast gate's workspace clippy runs with `--all-targets`, so test targets compile per change. The lint job's wall clock on this pull request is the measurement for the 10-minute budget.

`actionlint` is clean on every touched workflow.

Fixes #923


lexoliu:
Measured on this pull request's run: `Test / lint` with `--all-targets` took 7.7 min against a restored cache that carried no test-target artifacts (the last five dev lints were 3–5 min with `--lib --bins`); the gate's other leg, `Android FFI`, took 7.6 min, so the gate stays at its previous critical path and inside the 10-minute budget. The dev push after merge writes the cache with the test targets, so pull requests after it restore them.

PR #947, opened 2026-09-15
fix(badge): compose Badge in Rust instead of an unrealized native leaf

### LINKED ISSUE #928: Badge panics at launch: configurable!/Native leaf with no backend renderer or Rust fallback
`waterui::component::badge::Badge` is declared via `configurable!`, which makes it a `Native<BadgeConfig>` leaf requiring a backend renderer. No Apple renderer exists (`Sources/WaterUI` has no Badge component and the FFI header exports no `WuiBadgeConfig`), and `Native::body` panics when neither a renderer nor a fallback is registered:

```
panic.location="core/src/components/native.rs:66:32"
panic.payload="Native view (waterui_internal::component::badge::BadgeConfig)"
```

Captured by nightly e2e (apple-backend run 34924228215): the `reply` example packages fine but panics at launch on macOS the first time `material_badge` (hydrolysis-m3 → `WaterBadge`) is rendered.

This also contradicts the framework design rule that `Badge` is a Rust-side composer shipping zero new C-ABI types — it should compose `vstack`/`hstack`/overlay + theme tokens, not require a native leaf.

## Fix

Give `Badge` a Rust-side realization — either a pure composer `body` or `Native::new(config).with_fallback(...)` following the `table.rs`/`list` pattern — so it renders on every backend without new FFI surface.

lexoliu:
## Summary

- `Badge` was a `configurable!` component → `View::body` fell back to `Native<BadgeConfig>`, which panics on backends without a `BadgeConfig` realization. The packaged `reply` macOS app crashed at launch with `Native view (waterui_internal::component::badge::BadgeConfig)` (nightly run 34924228215).
- Replaced it with the documented design (AGENTS principle 2 lists `Badge` among the Rust-side composers): an `overlay` of a `when`-switched indicator over the content — positive counts render a bold-caption capsule, non-positive counts render a 6pt dot.
- Default indicator color now resolves the theme `Accent` instead of `Color::default()` (opaque black). The black default was previously invisible because hydrolysis' native `BadgeConfig` realization ignored `config.color`; it would have surfaced the moment the composer ran anywhere.

Fixes #928

#### Test plan

- [x] `cargo check -p waterui`, `cargo clippy -p waterui`, `cargo fmt --check` clean
- [x] `tests/e2e_composers.rs::badge_value_tracks_its_signal` still valid — the composer renders the count as a reactive `text!` label
- [ ] CI


lexoliu:
Closing: superseded by the native-leaf direction landed in #927 (`WuiBadge` FFI export) — converting Badge to a Rust composer would bypass hydrolysis' existing native BadgeConfig realization and dead-code the new FFI surface. The Apple-side fix for #928 is a WuiBadge renderer in water-rs/apple-backend instead.

Issue #1056, opened 2026-09-17
Freeze the layout contract at 0.5.0: last placement-proposal change, normative spec


lexoliu:
0.5.0 is the last release at which the core layout semantics may change. Before it ships:

1. One last verified divergence from the reference model is fixed: `HStack`/`VStack` keep the measurement probe as the cross-axis placement proposal, so a column widened past its proposal by an unshrinkable row still wraps its title at the proposal width. SwiftUI proposes the stack's resolved cross extent at placement (the Apple parity capture of `examples/hover` shows the title on one line across the 414 pt column while ours wraps to two lines at 370 pt). The rule becomes: main axis keeps the measured proposal's allocation, cross axis is the cross extent of the bounds the stack was placed in.

   Verified not to be core problems, so they stay as backend work: the `multi_window` cards (a headless Hydrolysis probe shows core stacks content-sized — a card without buttons measures 140 pt in a 362 pt column; the Apple width difference is the automatic button style's (8, 4) content padding) and scroll content wider than the viewport being leading-parked (the Apple `WuiScroll` framing; the contract is that scroll content is centred on the non-scrolling axis).

2. The whole contract — proposal protocol, stretch, stack sizing and allocation, every container, the leaf contracts backends implement, backend placement — is written down as `docs/layout-spec.md`, and `AGENTS.md` records that the layout is frozen and points at it. After this lands, a change to these semantics needs a major-version decision by the maintainer.

## Amended after the external review of the freeze proposal

The cross-axis-only rule in item 1 was reviewed against the reference model and found incoherent: re-proposing the bounds on one axis while replaying the measurement allocation on the other leaves a stretched stack's children with stale main-axis geometry (an aspect-ratio child resolved against the old cross extent, a `Frame` measured under a proposal it never places with). The rule that lands is the full one: placement is a fresh negotiation against the resolved bounds on **both** axes, for every container.

- Stacks allocate the main axis from `bounds.main` and propose `bounds.cross`; the cross envelope is `max(guide) + max(extent − guide)` with explicit guides unclamped and honoured on every alignment (Leading/Center/Trailing anchor the envelope in the bounds; custom alignments anchor leading).
- `ZStack`, `Overlay`, `Background`, `Padding`, `Absolute` propose the bounds (less insets) they place into and never clamp a child: an oversized child overflows.
- `Frame` measures its child under the exact proposal it places with, so its answer and its placement agree.
[...]

Issue #1014, opened 2026-09-16
0.5.0 release: nightly green, satellite wave on crates.io, tagged backends, user-path e2e


lexoliu:
Parent for the 0.5.0 release. Carries the intent and groups the work; no implementation hangs off it. Release bar (2026-09-15/16):

1. Every known bug has a verification result and is closed, labelled or not; already-fixed ones are closed with the fixing PR named.
2. No `[patch]` or git reference to a water-rs repository remains in the tree that publishes `waterui` 0.5.0. Patches on third-party crates we forked (`lexoliu/vello` for the Mali device loss linebender/vello#1817 and the GL atlas fix #1859, `rust-block`, `rust-xcb`) stay as workspace-only patches until upstream ships; `cargo publish` ignores them, and the release notes say which fixes are dev-only.
3. Each repository's CI proves the user path on the same candidate inputs: install the published CLI, `water create`, debug run, `water package`, and the packaged artifact runs.

Mechanics: satellites (hydrolysis, hydrolysis-m3, gtk, dew, svg, math, chart, map-gpu, image, canvas, barcode, particle, video-gpu, visualizer) can only publish against a published 0.5.0 foundation, and the workspace can only compile against satellites that carry the 0.5.0 API, so the framework merges to `main` twice: the first run publishes the foundation and stops at the first crate whose satellite requirement is not on crates.io yet; the satellites release; a cleanup PR removes the git patches; the second run publishes `waterui-testing`, `waterui-media`, `waterui-internal`, `waterui-dylib`, `waterui-inspector-app` and the `waterui` facade at 0.5.0 from a git-free tree, which is where `v0.5.0` and `framework.json` land. Backends: apple-backend 0.3.0 and android-backend 0.3.0 tags; `waterui-cli` 0.3.2 (cli#30) is the floor `minimum-cli-version` already declares.

Enhancements previously on this milestone moved to 0.6.0; the milestone is the release checklist.

lexoliu:
0.5.0 is out. Evidence per part:

| Part | Evidence |
| --- | --- |
| Nightly green | waterui nightly certified and promoted; sub-issues #950 #1010 #1011 #1012 #1013 closed with their runs. |
[...]

PR #1057, opened 2026-09-17
docs: freeze the layout specification at 0.5.0

lexoliu:
docs: freeze the layout specification at 0.5.0

`docs/layout-spec.md` is the normative description of the layout protocol:
proposals and the three-point contract, `SubView` and `Layout`, stretch,
stack distribution (one probe, minimum/ideal, priority bands, water-fill)
and placement (children are placed with the proposal they were measured
with — a deliberate difference from SwiftUI, which re-proposes the placed
bounds), every container's rule, and the leaf contracts backends implement
(text answers its laid-out width, a spacer its minimum length, a scroll
claims the offer and centres its content on the non-scrolling axis).

AGENTS.md gains Principle 11: the layout system is frozen at 0.5.0; a pull
request that changes a documented behaviour or weakens a layout contract
test is rejected, and a backend that disagrees with the specification is
fixed on that backend. The engagement rules point layout diagnosis at the
document first.

Part of #1014


lexoliu:
Superseded: the layout freeze lands as one PR carrying the spec, the AGENTS.md conclusion and the single approved core change (Fixes #1056) once the external review of the proposal returns; the Principle 11 wording from this branch is folded into it.

PR #1107, opened 2026-09-18
build: pin waterui-map-gpu at 11cac93 (raster job joins the local task set)

### LINKED ISSUE #25: Raster tile job is invisible to hosts that settle on monitored local tasks
## Problem

The tiled raster job that turns prepared vector display lists into map tile
textures is spawned detached on the global executor:

```rust
// src/render.rs, start_raster_job
#[cfg(not(target_arch = "wasm32"))]
executor_core::spawn(async move {
    let finished = blocking::unblock(move || {
        std::panic::catch_unwind(...)
    })
    .await;
    ...
    redraw_handle.request_redraw();
})
.detach();
```

Work spawned this way is invisible to
`waterui::task::outstanding_local_tasks`, which is the completion signal
frame-driven hosts (the `water preview` headless runtime, `waterui-testing`)
pace on before they report a frame settled. Sequence on a non-GL backend
(`gpu_calls_thread_bound == false`, e.g. Metal):

1. `MapRequestTask` (a counted `spawn_local`) fetches style + tiles, writes
   `pending_prepared`, finishes — outstanding count drops to 0.
2. The next render activates the prepared map and `start_raster_job` detaches
   the raster job onto the global executor.
3. The host's settle loop sees a quiet tree and zero outstanding local tasks,
   returns, and captures — while the raster job is still running.
4. The snapshot shows the map surface without its tiles.

The job completes ~150–200 ms later and calls `request_redraw`, but by then
the frame was already captured. The wasm path does not have this bug: it runs
the same job through `spawn_local`, which *is* counted.

## Proposed fix

Run the non-wasm raster job through `spawn_local` as well, keeping
`blocking::unblock` for the blocking rasterization itself:

```rust
spawn_local(async move {
    let worker_sender = sender.clone();
    let finished = blocking::unblock(move || {
        std::panic::catch_unwind(core::panic::AssertUnwindSafe(|| {
            drive_raster_job(job, &worker_sender, true, &notify);
        }))
        .is_ok()
    })
    .await;
    ...
    redraw_handle.request_redraw();
})
.detach();
```

The `spawn_local` future stays parked on the `unblock` join for the whole
job, so `outstanding_local_tasks` stays non-zero until the raster publishes —
the same contract `MapRequestTask` already follows. The wasm branch is
unchanged.

## Reproduction

```
water preview demo --backend hydrolysis --theme material3 \
  --frame 900x640 --path examples/map --output map.png
```

in the `waterui` workspace (requires `water-rs/cli` ≥ the fix for
water-rs/cli#93, which makes the preview runtime build its environment through
the app's composition root so `waterui_map_gpu::install` runs at all). The
captured frame shows the map chrome and a blank tile area; adding a wall-clock
pacing loop after settle makes the tiles appear.

lexoliu:
## Summary
- Moves the `waterui-map-gpu` patch rev to 11cac93 — the squash merge of water-rs/map-gpu#26, which fixes water-rs/map-gpu#25.
- The raster job was a detached `executor_core::spawn` task, invisible to `outstanding_local_tasks`; Hydrolysis's settle loop captured the frame before tiles published. `water preview` on the map example produced map chrome over a blank tile surface.
- The job now runs as a local task awaiting `blocking::unblock`, so settle waits for the raster to publish while the work still runs on the blocking pool.

#### Test plan
- [x] `water preview --backend hydrolysis` on `examples/map` renders full Manhattan tiles with the generated backend patched to this rev
- [x] map-gpu CI green on the merge commit


lexoliu:
Superseded by #1110, which pins `waterui-map-gpu` at `fa83ba7b` — map-gpu's current dev head, which contains this PR's target `11cac93` (the raster-job-as-local-task fix) plus water-rs/map-gpu#28. Closing; the map preview fix is on dev through that pin.

PR #1118, opened 2026-09-19
ci: cache prebuilt tool installs so a CDN failure cannot kill a leg

### LINKED ISSUE #1116: Tool installs in CI are a single point of failure: one CDN 504 kills a 40-minute leg
## Defect

Every CI job that installs a tool over the network can be killed by a transient download failure after it has already paid for checkout, toolchain setup, and cache restore.

Nightly run `35347280479` (2026-09-18), job `Test / macos-latest`, failed exactly this way — eleven consecutive retries against a CDN that was serving 504s, then the step exited and the whole leg went red:

```
curl: (56) The requested URL returned error: 504   (x11, 12:56:21 -> 12:57:18)
##[error]Process completed with exit code 56.
```

Nothing in the repository was broken; a tool download returned HTTP 504. The leg had already paid for checkout, toolchain and cache restore.

## Exposure

Every `taiki-e/install-action` use shares it — `test.yml` installs nextest in four jobs (`test`, `all-features`, `macos-suites`, `examples`), `bench.yml` in the bench matrix, `ci.yml` installs cargo-audit / cargo-deny / cargo-machete in the hygiene job, `dep-check.yml` installs cargo-edit, `build-cli.yml` installs cargo-dist twice, and `release-preflight.yml` installs cargo-binstall plus the registry tooling. The same class covers `crate-ci/typos` (downloads the typos binary every run) and the `dxc` release-asset download in `install-dxc`. install-action retries the download but has no "already installed" path, so a warm run hits the network every single time and a CDN outage of more than a minute is fatal.

## Fix direction

Make a transient download failure survivable: pin the tool versions install-action resolves (`tool: name@version`, its documented input), and cache the installed binaries under a key derived from that spec so a healthy run restores them and never touches the network. `actions/cache` only saves on `success()`, so an interrupted install cannot poison the entry. This is a robustness and cost change only — coverage, platforms, feature combinations and lint gates are unchanged.

lexoliu:
Fixes #1116

## Summary

Nightly run 35347280479 lost `Test / macos-latest` to eleven retries against a CDN serving 504s — the step exited and a 40-minute leg went red with nothing wrong in the repository. `taiki-e/install-action` retries downloads but has no already-installed path, so every run hit the network.

- New `.github/actions/install-tools` composite wraps `taiki-e/install-action` behind an `actions/cache` entry whose key is derived from the exact `tool` spec. Every call site now pins the version via install-action's documented `name@version` input, which is what makes the key exact. A warm run restores the binaries and never touches the network; `actions/cache` only saves on `success()`, so an interrupted install cannot poison the entry.
- The composite caches both directories install-action can install into (`~/.cargo/bin` when a rustup cargo is present, `~/.install-action/bin` otherwise) and re-exposes the fallback dir on a cache hit.
- The same restore-and-gate pattern covers the sites that do not go through install-action: the `dxc` release-asset download (in `install-dxc`), the `cargo-local-registry`/`margo` tooling in release-preflight, and the `uv` binary (tool-cache directory + pinned version).
- `typos` moves from `crate-ci/typos@v1.49.0` to `typos-cli@1.49.0` through install-tools — identical tool version, now cached. `oven-sh/setup-bun` already restores its binary from a cache keyed by its download URL, so it needed no change. `nttld/setup-ndk` keeps its documented `local-cache: false`. `cargo install waterui-cli --version ">=$floor"` stays uncached: the install is the check, not provisioning.

Install sites changed:

| File | Site | Tool spec |
| --- | --- | --- |
| test.yml | 4 jobs (test, all-features, macos-suites, examples) | `cargo-nextest@0.9.145` via `NEXTEST_VERSION` env |
| nightly.yml | coverage | `cargo-llvm-cov@0.9.1`, `cargo-nextest@0.9.145` |
| nightly.yml | feature-checks | `cargo-hack@0.6.45` |
| windows.yml | test, all-features | `cargo-nextest@0.9.145` |
| windows.yml | 3 jobs | `install-dxc` now caches the release archive |
| ci.yml | hygiene | `cargo-audit@0.22.2`, `cargo-deny@0.20.2`, `cargo-machete@0.9.2` in one call |
| ci.yml | typos | `typos-cli@1.49.0` + `run: typos` |
| ci.yml | scripts | `uv` pinned to 0.12.17 behind a tool-cache restore |
| dep-check.yml | check | `cargo-edit@0.13.13` (documented `cargo-binstall` fallback, same as today) |
| release-preflight.yml | resolve-gate | `cargo-local-registry@0.2.12`, `margo@0.1.7` under one entry |

What a 504 now does: on a warm run nothing is downloaded at all — the binaries restore from cache and the job proceeds. On a cold run (first run for a spec, or a version bump) the download still has to happen and can still fail; `actions/cache`'s `post-if: success()` means that failure writes no cache entry, so the next run simply retries rather than inheriting a poisoned restore. No check was weakened: same tools, same versions install-action resolves today, same commands afterwards.

## Test plan

- [x] `actionlint` clean across workflows and the new composite action
- [ ] CI green on this pull request


lexoliu:
Superseded by #1121, which covers the same install sites plus browser-runtime-wpe.yml, ffi-header-cache.yml and refresh-apt-index, and fixes the release-preflight baseline fetch that reported exists=false on a transport error.

